PR #7, opened 2026-01-28
Add an MSRV policy; test against the MSRV.

jrvanwhy:
This adds an MSRV policy to tock-registers. This policy currently matches libtock-rs: bump to a newer stable release anytime doing so is convenient. This MSRV is specified in Cargo.toml and, when combined with #6, tests it in CI.

1.82 is the earlier stable Rust version that can run `cargo clippy` without warnings (earlier version give the error clippy::unused_result_ok). Note that 1.82 does *not* support the 2024 edition, if we want to update that then we should switch to 1.85.

### DIFF
--- a/.github/workflows/license_check.yaml
+++ b/.github/workflows/license_check.yaml
@@ -22,5 +22,5 @@ jobs:
           repository: tock/tock
 
       - name: License check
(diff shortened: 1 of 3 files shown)


brghena (approved):
Looks good to me. Updating to a newer version of Rust can be spun into other updates to the system.

Issue #4, opened 2026-01-13
References to MMIO space are (probably?) UB


thejpster:
I was looking at tock-registers in detail for the first time today, and I think you have the same soundness issue that we fell over in the svd2rust generated output.

As far as we (me, and others in the Embedded Devices WG) can tell, it is UB to have a reference which points to MMIO address space. I believe tock-registers generates such MMIO references.

The UB comes about because references in Rust are marked as dereferencable when handed to LLVM, and that means LLVM is allowed to read from them whenever it likes. This would be incorrect for a reference to MMIO space, because the read may have side-effects. I think currently LLVM does not actually do any unexpected dereferencing, but it remains allowable.

I believe the only correct approach is to only ever construct pointers to MMIO space, and not references, and be very careful to never accidentally conjure up a reference. This is the approach taken by [safe-mmio](https://crates.io/crates/safe-mmio) and [derive-mmio](https://crates.io/crates/derive-mmio), and is what svd2rust is working towards.

thejpster:
I think https://github.com/rust-embedded/wg/issues/791 has most of the details, and links to bits of history.

brghena:
Thanks!! I believe we're aware of this issue, although revising the implementation of tock-registers has become a rather long-term endeavor. It's the first item @jrvanwhy notes in the draft he made discussing how to start revising tock-registers, which is still over on the main Tock repo: https://github.com/tock/tock/pull/4001

More references to the issue and pointers to community solutions to it is super helpful though.

jrvanwhy:
Yes, this is the soundness issue I've been wanting to fix for years, and yes tock/tock#4001 is the most recent attempt to alleviate it.

One thing that is delaying the fix is that I'm trying to fix a second design issue at the same time: the fact that tock-registers does nothing to make code that uses it unit-testable. I'm trying to fix both in a big monolithic redesign, and that is quite a bit of effort. To my knowledge, the only Rust registers library that is unit testable is Caliptra's ureg... maybe. I haven't fully figured out how it works to confirm that it is easy to unit test drivers built on ureg. Even if ureg does support that, we'll almost certainly implement our own because the APIs are too different (we have enough code that it is easier to redesign tock-registers than to migrate to a drastically-different API).

I've recently (as of 4 days ago!) resumed working on the redesign. My past attempt petered out as a result of team changes/reprioritization at my previous employer; I have solid reason to believe that this time will be different and we'll be able to get the fix/redesign across the finish line.

[...]

PR #11, opened 2026-04-01
Try 3 at sound and testable register definitions.

jrvanwhy:
# PR Context

This PR is the Tock project's plan for fixing #4 and adding testing support to `tock_registers`. We are currently merging this design into tock-registers (as a series of smaller PRs).

# PR Description

This PR introduces `register_map!` to `tock_registers`. `register_map!` is a successor to `register_structs!` that makes several improvements:

1. It resolves the soundness issue described in https://github.com/tock/tock-registers/issues/4, by using raw pointers to refer to registers rather than references.
1. It supports driver unit tests. In addition to a `struct`, it emits a trait representing the registers, which can be used by unit tests to simulate the hardware's behavior.
1. It better supports registers with non-pointer-sized addresses, such as x86 IO ports (which are addressed by `u16`s).
1. It allows external crates to define new operations (to e.g. support SPI peripherals with registers that support bitwise operations). This also makes it easy to support unsafe read/write operations, as described in https://github.com/tock/tock/issues/3244.

There are some feature requests that this does NOT address:

1. `#[cfg()]` attributes on registers (https://github.com/tock/tock/issues/3257). I don't think that merely passing `#[cfg()]` attributes through to the generated code is the correct solution for that use case, as the unit test environment will be impacted by those cfg attributes as well. Instead, I think that `register_map!` should support different device variants more natively, such that it always generates all the variants (so that the unit test environment can test every variant). Regardless, that seems like quite a bit of work, so its design and implementation are left out of this PR.
1. Typestate-based modeling of hardware states (from TockWorld 8). However, it is possible for a separate typestate library to integrate with `register_map!` to create registers that can only be accessed through the typestate struct (see [`examples/typestate.rs`](https://github.com/jrvanwhy/tock-registers/blob/mod-interface/examples/typestate.rs)).
1. DMA fences, though see [`examples/safe_dma.rs`](https://github.com/jrvanwhy/tock-registers/blob/mod-interface/examples/safe_dma.rs) for a demonstration of how an external crate could provide a mostly-safe DMA interface on top of `register_map!`.
1. RISC-V CSR support. Supporting registers whose addresses are hardcoded inside instructions complicates the design significantly. I realized that the amount of code and complexity that I would have to add to tock-registers to support CSRs would exceed the size of the entire riscv_csr crate, so I decided to omit that functionality.

This has the following advantages relative to my previous attempt to replace `register_structs!` (tock/tock#4001):

1. It supports arrays of register blocks for peripherals with repeating patterns of registers.
1. It supports nesting register blocks inside other register blocks (for e.g. peripherals that share a subset of registers, peripherals that are extended versions of other peripherals, etc).
1. It supports non-MMIO registers, such as x86 IO ports, and allows external crates to add new register types.
1. Offset tests are run more consistently: instead of only being run when the register block is compiled for the target system, all offset tests are run on every `cargo check`.

# Review guide

I suggest starting with:

1. The [new README](https://github.com/jrvanwhy/tock-registers/blob/mod-interface/README.md)
1. The [documentation](https://github.com/jrvanwhy/tock-registers/blob/mod-interface/src/map.rs) for the `register_map!` macro (this is much more normative and complete than the README documentation, but less approachable).
1. The [doc/](https://github.com/jrvanwhy/tock-registers/tree/mod-interface/doc) directory.

When you want usage examples, you can also look at:

1. The [tests/](https://github.com/jrvanwhy/tock-registers/blob/mod-interface/tests) directory.
1. The [examples/](https://github.com/jrvanwhy/tock-registers/blob/mod-interface/examples) directory.
1. The [ast module](https://github.com/jrvanwhy/tock-registers/blob/mod-interface/codegen/src/ast.rs). This is especially useful if you want to look at the parsing code.

If you want to see the full generated code, take a look at:

1. [single_test_scalar](https://github.com/jrvanwhy/tock-registers/blob/mod-interface/codegen/src/single_test_scalar.rs)
1. [single_test_array](https://github.com/jrvanwhy/tock-registers/blob/mod-interface/codegen/src/single_test_array.rs)
1. [block_test_all_fields](https://github.com/jrvanwhy/tock-registers/blob/mod-interface/codegen/src/block_test_all_fields.rs)

At this point, this entire PR is ready for review and discussion.

There is a companion PR at https://github.com/tock/tock/pull/4772 that ports a couple of Tock drivers to this API.


src/map.rs
jrvanwhy, line 1: # Bikeshed: what should this macro be called

`registers!` is really the core feature here, but I am unhappy with the name. There is already a `tock_registers::registers` module in tock-registers, plus this PR adds a `Register` trait. I'd rather have a different, more distinct name for the macro.

And so far, I have not come up with a name I like. In tock/tock#4001, I solved this by calling the macro `peripheral!`, but this PR's macro also supports definining individual registers, so that name doesn't really fit. I could be consistent with `register_structs!` and call it `register_mods!`, perhaps?

Discussing this now because handling the rename will be a really big find-and-replace.

src/fake_register.rs
```diff
@@ -0,0 +1,190 @@
...
+            write: (),
+        }
+    }
+}
+
+impl<Data: Copy, DT: DataType, R: Access, W: Access> FakeRegister<Data, DT, R, W> {
+    /// Returns a new FakeRegister that implements [`trait@Read`] by invoking `fcn`.
+    pub const fn on_read(
+        self,
+        fcn: fn(Data) -> LocalRegisterCopy<DT::Value, DT::LongName>,
```
jrvanwhy: I'm conflicted on whether the function signatures for `on_read`, `on_write`, `on_unsafe_read`, and `on_unsafe_write` should use `LocalRegisterCopy` or simply `DT::Value`.

I would expect fake peripherals to want to use LocalRegisterCopy on registers that have bitfields and not use LocalRegisterCopy on registers without bitfields. Unfortunately, I don't see a way to make FakeRegister easily support both.

I did some grepping in the `chips/` directory, and found:

* 1011 primitive registers
* 3812 bitfield registers

Based on that, I chose to use LocalRegisterCopy, but I'm not confident in that choice.

README.md
brghena, line 440: Is this performance expectation in the README still valid? "everything compiles
down to the optimal inlined bit twiddling instructions--in other words, there is zero runtime cost"

brghena, line 4: We should clearly document which version this README points to. Is this going to be version 2.0 of Tock registers? I know we never hit 1.0, but I think it might make sense to signal that this is a whole not-compatible but significantly better thing.

src/map.rs
brghena: `register_mods!` makes sense to me. Or maybe `register_definitions!` or `register_layouts!` to contrast with `register_bitfields!`

src/map.rs
```diff
@@ -0,0 +1,349 @@
...
+///     #[buses(Mmio32, Mmio64)]
+///     dma_rng {
+///         0 => address: usize { Read, Write },
+///         [4, 8] => length: u32 { Read, Write },
+///         // Padding can also move around, and have per-bus length.
+///         [8, 12] => _: [4, 0],
+///         // Because of the previous padding, this is at offset 12 on both buses.
+///         12 => status: u8 { Read },
+///     }
+/// }
```
brghena, line 249: Oof, this would be really hard to read and understand. Is it possible, or better, to just make two separate register layouts for LiteX and choose between them at compile time?

doc/UnitTesting.md
```diff
@@ -0,0 +1,95 @@
...
+pub fn getrandom_real(buffer: &mut [u8]) {
+    use { core::ptr::without_provenance_mut, rng::Interface as _ };
+    // Safety: We know this CPU has the RNG peripheral at address 0x10000.
+    let registers = unsafe { rng::Real::<Mmio32>::new(Mmio32(without_provenance_mut(0x10000))) };
+    getrandom(registers, buffer);
+}
+```
+
+Now code that requires the real hardware can use `getrandom_real`, while code
+that needs to remain unit-testable can use `getrandom`.
```
brghena: I don't understand this part. Why do we need both `getrandom_real()` and `getrandom()`?

doc/UnitTesting.md
```diff
@@ -0,0 +1,95 @@
...
+    #[test]
+    fn getrandom_test() {
+        let mut buffer = [0; 3];
+        getrandom(&FakeRng::default(), &mut buffer);
+        assert_eq!(buffer, [1, 2, 3]);
+    }
+}
+```
+
+See [examples/rng.rs] for the complete example.
```
brghena: Broken link

brghena (commented):
I only looked at documentation, not implementation details, but this makes sense to me. Mostly my opinion is "the external view of this looks pretty similar to how it used to look, great". I recorded some comments as I read through the documentation.

Requiring the listing of offset values is a strictly positive change and helps prevent mistakes. Bitfields seem to work exactly the same, so that's excellent. And adding `()` to all the register names is annoying but harmless.

If you tell me that 1) this is actually sound and 2) that it has the same performance guarantees, then I'm quite happy with it.

README.md
jrvanwhy: I have verified that https://github.com/tock/tock/pull/4772 has zero impact on the size of the generated binaries for the two boards that I changed (one is RISC-V, one is x86). When I make a change that I think is likely to impact optimizations (such as when I migrated Mmio32/Mmio64 from `*mut ()` to NonNull), I re-check that, but haven't seen a change yet.

I have not implemented benchmarks to compare this code against hand-optimized bit twiddling or verified that the generated assembly is the same.

So I believe the README statement is still valid, but I have not spent a large amount of effort demonstrating that.

brghena: That's a good enough comparison for me.

jrvanwhy: Leon and I have different ideas about what version number this will become, and discussed bringing it up at the core WG call.

I've been treating this README as documenting the current state of the repository, which means that it may reflect an unreleased version. I'm open to us changing it to reflect something else.

I also plan to update the crate-wide comment to reflect the new API, and I think it should be pretty similar to README.md. That comment shows up when you go to docs.rs/tock-registers and reflects the latest version (or whichever version you have selected in the docs.rs interface).

src/map.rs
jrvanwhy: Do you mean that the macro would generate two completely separate modules? Then you can't write and unit test a driver that works on both without leaning way too hard on `#[cfg]`.

Or do you mean you'd write:

```rust
registers! {
    #[bus(Mmio32)]
    dma_rng {
        0 => address: usize { Read, Write },
        4 => length: u32 { Read, Write },
        8 => _: 4,
        12 => status: u8 { Read },
    },

    #[bus(Mmio64)]
    dma_rng {
        0 => address: usize { Read, Write },
        8 => length: u32 { Read, Write },
        12 => status: u8 { Read },
    },
}
```

and the macro would detect the duplicate definitions, verify that they are in fact equivalent, and generate a single module that performs both.

That's technically possible but seems really messy and would require a lot of duplication. Also, what do you do with the doc comments -- duplicate them? Only have them on one definition? Maybe we'd have a "register list" syntax and a separate "offset specification" syntax, so you have a single register list with multiple offset specifications.

doc/UnitTesting.md
jrvanwhy: We need `getrandom()` for unit-testability, but ultimately the driver does not need to expose `getrandom_real()`.

In case it's unclear, at this point in the documentation the public interface is:

```rust
/// Unit-testable driver implementation.
pub fn getrandom<R: rng::Interface>(registers: R, buffer: &mut [u8]) { ... }

/// Implementation that talks directly to the hardware.
#[cfg(target = <whatever>)]
pub fn getrandom_real(buffer: &mut [u8]) { ... }
```

In the context of Tock, we don't really need `getrandom_real`. For us, our driver is likely to be a generic struct that looks like:

```rust
pub struct RngDriver<R: rng::Interface> { ... }
```

When compiled for the target system, the chip crate will instantiate `RngDriver<rng::Real<Mmio32>>`. When compiled for a unit test environment, the unit tests will use other instantiations.

If we decide to split chips into `forbid(unsafe_code)` and `allow(unsafe_code)` crates, then the definition and unit tests for `RngDriver` would be in the `forbid(unsafe_code)` crate while the code that instantiates `RngDriver<rng::Real<Mmio32>>` will be in the `allow(unsafe_code)` crate.

But this documentation is not Tock-specific, so it doesn't exactly match the Tock project's code patterns.

brghena: I guess my problem is with the wording in the documentation then.

The first snippet has a `get_random()` implementation that uses the registers directly. I think leave that part alone.

The second snippet creates a new testable function. We should call that `get_random_testable<R: rng::Interface>(...`.

Then in the last part, you could say:

---


Code that wants to be testable but needs real register instantiation for hardware can call the testable function with a wrapper that first instantiates the registers.

```
/// Fills the provided buffer with random / data.
pub fn get_random(buffer: &mut [u8]) {
    use { core::ptr::without_provenance_mut, rng::Interface as _ };
    // Safety: We know this CPU has the RNG peripheral at address 0x10000.
    let registers = unsafe { rng::Real::<Mmio32>::new(Mmio32(without_provenance_mut(0x10000))) };
    get_random_testable(registers, buffer);
}
```

---

I think that would make more sense to me. You're just demonstrating that a wrapper could exist.

Alternatively, I don't think it would be unreasonable to show a struct holding on to the registers as a field that member functions could access is unreasonable, as that achieves both testability and usability really well.

brghena: Thinking about my comment above, I think something triggering to me is the idea that we have to litter our codebase with `_real` so that unit tests are possible. I don't think that's actually the case at all, but it felt to me like what the original code was showing until you explained more, that to access hardware we'd need a separate `getrandom_real()` function.

src/map.rs
brghena: I think I meant separate modules. But really I'm asking if there's any good other way to achieve the goal.

The "macro would detect the duplicate definitions" idea seems like a non-solution given the complexity.

My concern is that this "you can have a list of busses in order" solution is also too complex, maybe not in internal implementation but in external use.

For the separate modules idea, the extension of that would be two LiteX implementations, one for 32-bit and one for 64-bit that share some functionality. I think that might be reasonable to expect rather than having one 32/64 bit implementation that's too hard to reason about. You don't lean hard on `cfg`, you just make your peace with writing it twice when there are two differing versions.

This is just my opinion though, and I haven't necessarily thought through the ramifications of that.

README.md
brghena: > I've been treating this README as documenting the current state of the repository, which means that it may reflect an unreleased version. 

Good point. Yes, I agree with that. But I think there should be at least a notification sentence up there about it being a "brand-new Tock registers".

src/map.rs
jrvanwhy: For drivers that only want to support a single bus type (i.e. almost all non-LiteX drivers), the only extra complexity that they see is:

1. The bus specification attribute is `#[buses()]` instead of `#[bus]` (and in practice, that's hidden by `mmio32_registers!` and `mmio64_registers!` anyway).
2. The generated module contains a `Bus` trait
3. The generated module's Real struct has a `<B: Bus>` generic argument.

@lschuermann: The whole point of your LiteX register abstraction types is to avoid requiring drivers for the same LiteX peripheral to be duplicated between chips with different bus configurations, correct?

Because this seems like a desirable feature, and it only adds significant complexity for use cases that are using it.

brghena: Things I definitely agree with:
* Make complicated things possible
* Make sure simple things aren't made significantly worse by that
* Not giving you more work, especially for a niche case
* Generally, avoiding duplication if possible

I think this feature does all of that stuff well.

The only thing I'm questioning is if this is usable enough to support the LiteX case. My personal opinion is maybe no, and that we should explore other methods for the LiteX case. But I am very happy to hear from @lschuermann on that.

jrvanwhy: I've just spent about an hour workshopping this with Leon.

Ultimately, support for multiple buses involves a number of bits of complexity:

1. Naming the buses specification attribute `#[buses]` rather than `#[bus]`.
2. Emitting a Bus trait.
3. Adding a `<B: Bus>` generic argument to the Real struct.
4. Specifying multiple buses in the `#[buses]` attribute.
5. Specifying multiple offsets and/or padding sizes in the register block definition.
6. Maybe: some references to Bus/Address that can be removed in traits (I'm not sure if any of these are removable though, I suspect not).

1-3 are from my last comment, and 4-6 are pieces of complexity that are either internal to tock-registers or which are only seen by drivers that use multiple-bus support.

The one bit of complexity I really want to get rid of is the `<B: Bus>` generic, because that's the one that shows up everywhere you try to use tock-registers.

Currently, the design uses the `<B: Bus>` generic to support a number of use cases:

1. LiteX
2. BorrowedBus, which is used to make tock-registers safe in symmetric multi-processing environments (see RegisterSender)
3. Peripherals that are used on both CHERI and non-CHERI platforms (this one is a bit nonobvious, but to support CHERI we'll have to add a register type that can hold CapabilityPtr on CHERI systems and a normal pointer on no-CHERI systems, which requires separate buses to get the offsets correct).
4. The ARMv6 interrupt vector table, which is sometimes at 0 (Mmio32Nullable) and sometimes at other addresses (Mmio32).
5. 16550 UARTs, which have port IO and MMIO implementations (with different MMIO maps on different systems).

And probably more we haven't thought of.

Most of these are individually very niche, except `BorrowedBus`, which is really not niche. Ultimately, I think that use case alone justifies having a `<B: Bus>` generic on the `Real` struct. The fact that a bunch of other niche use cases also benefit from that is a bonus.

The pieces of complexity that `BorrowedBus` requires either do not affect single-bus driver developers (4, 5) or have very minimal complexity impact (1, 6).

So I think it makes a lot of sense to keep this functionality even if we completely disregard LiteX.

jrvanwhy: Hmm, we could consider adding a `#[bus]` attribute that works exactly the same as `#[buses]` except:

1. It only accepts a single bus specification
2. It changes the Real generic to `<B: Bus = #bus>` where `#bus` is the specified bus.

That would help remove some of the verbosity from user code, as you could specify `Real` rather than `Real<Mmio32>` in most places.

We'd want to switch `mmio32_registers!` and `mmio64_registers!` to use `#[bus]` instead of `#[buses]`. I believe that would be a backwards-compatible change so we can do it later.

lschuermann: Thank you @jrvanwhy for summarizing this conversation, and my apologies for my late response. I agree with the summary, but I want to emphasize for clarity (as I thought that I did in the past already): **we should not use LiteX alone to justify any additional complexity added to tock-registers**.

The current hack LiteX uses for supporting multiple underlying buses is pretty cursed, so we'd want to get rid of it one way or the other. Furthermore, the only two bus configurations that LiteX uses in practice, except for extremely exotic targets, are 32-bit and 64-bit buses with 32-bit and 64-bit words respectively. We're not even supporting 64-bit, but even if we were, duplicating a registers definition wouldn't be a big deal either.

So, if we happen to support multiple different buses anyways, we may as well use that to retain this flexibility for the existing LiteX peripherals. But, for our deliberations, let's pretend that LiteX doesn't exist and go from there.

src/map.rs
jrvanwhy: I kinda like `register_layouts!` because of its similarity to `register_bitfields!`.

src/fake_register.rs
jrvanwhy: To be clear, the reason this isn't cut-and-dry is that choosing to use LocalRegisterCopy forces the fakes to use LocalRegisterCopy on every register, even if that register does not have bitfields. If FakeRegister used `DT::Value` instead, then only the registers that use bitfields would use LocalRegisterCopy.

README.md
jrvanwhy: Good point about the warning, I added it ([Rendered](https://github.com/jrvanwhy/tock-registers/tree/mod-interface)).

doc/UnitTesting.md
jrvanwhy: Okay, I've reworked this somewhat ([Rendered](https://github.com/jrvanwhy/tock-registers/blob/mod-interface/doc/UnitTesting.md)). Now the unit-testable function is called `getrandom_impl`, while the real implementation is just `getrandom`.

brghena: This is great! This flow makes way more sense to me.

doc/UnitTesting.md
brghena: Link is still broken for some reason

jrvanwhy: > Link is still broken for some reason

I know EXACTLY why it's broken: I fixed it wrong.

It should be fixed now. To make it as easy as possible to understand this PR and its review history, I'm avoiding force pushes and creating large numbers of small commits. Therefore, I wait to make little fixes like this until the next time I have a substantial commit to push.

jrvanwhy:
I have performed the following renames (recording this here so the PR discussion makes sense in the future):

1. `tock_registers::registers!` is now `tock_registers::register_layouts!`
1. `tock_registers::mmio32_registers!` is now `tock_registers::mmio32_register_layouts!`
1. `tock_registers::mmio64_registers!` is now `tock_registers::mmio64_register_layouts!`

src/map.rs
jrvanwhy: > Hmm, we could consider adding a `#[bus]` attribute that works exactly the same as `#[buses]` except:
> 
>     1. It only accepts a single bus specification
>     2. It changes the Real generic to `<B: Bus = #bus>` where `#bus` is the specified bus.
> 
> 
> That would help remove some of the verbosity from user code, as you could specify `Real` rather than `Real<Mmio32>` in most places.
> 
> We'd want to switch `mmio32_registers!` and `mmio64_registers!` to use `#[bus]` instead of `#[buses]`. I believe that would be a backwards-compatible change so we can do it later.

`#[bus]` is now implemented, and I updated tock/tock#4772 as well.

README.md
```diff
@@ -3,127 +3,116 @@
...
+Note: `register_layouts!` has several pieces of functionality that are not
+described in this document. See its Rustdoc for complete documentation.
+
+A register block is defined using the `register_layouts!` macro:
 
 ```rust
-use tock_registers::registers::{ReadOnly, ReadWrite, WriteOnly};
+use tock_registers::{register_layouts, Mmio32, Read, Write};
+
+register_layouts! {
```
bradjc: I'm not sure what the plural is here. Presumably each register has a single layout. I think I'm thinking of "register map" when I read this, where a single map covers the registers for a peripheral. In this case, the layout corresponds to the peripheral, not an individual register, making it singular.

README.md
```diff
@@ -3,127 +3,116 @@
...
+    // Specifies that these registers are memory-mapped IO registers with 32-bit
+    // addresses.
+    #![buses(Mmio32)]
```
bradjc: Is there a use case for a register layout supporting multiple buses?

README.md
```diff
@@ -3,127 +3,116 @@
...
-register_structs! {
-    pub Registers {
-        (0x000 => foo: ReadOnly<u32>),
-        (0x004 => pub bar: ReadOnly<u32>),
-        (0x008 => @END),
+use tock_registers::{mmio32_register_layouts, Read};
+
+// mmio32_register_layouts! is shorthand for register_layouts! with the Mmio32
+// bus. mmio64_register_layouts exists as well.
+mmio32_register_layouts! {
```
bradjc: On one hand, having multiple ways to do the same thing tends to fragment code and add complexity. On the other, this is arguably more intuitive because it doesn't introduce what a "bus" is, which chip developers largely don't need to consider.

I wonder if we should just use this for the example (since nearly everything will use it), and then later introduce the more general version.

bradjc (commented):
Looks pretty good to me!

A couple notes/questions on naming.

I would like there to be readmes for the crates which just make it easier for new readers to understand what the different crates are for.

Note, I did not go through the code in any depth.

README.md
jrvanwhy: Ah, yeah I definitely did not intend "register layout" to mean a similar thing to "register map".

In my terminology, a "layout" is basically any top-level thing inside "register_layouts!". We've already bikeshed the name somewhat, which resulted in the rename to `register_layouts!`. I'm fine with another rename if we come up with a better name (the rename was much easier than expected, so I'm happy to do it again).

README.md
jrvanwhy: 1. LiteX
1. Peripherals used on CHERI and non-CHERI platforms (this one is a bit nonobvious, but to support CHERI we'll have to add a register type that can hold CapabilityPtr on CHERI systems and a normal pointer on no-CHERI systems, which requires separate buses to get the offsets correct)
1. The ARMv6 interrupt vector table, which is sometimes at 0 (Mmio32Nullable) and sometimes at other addresses (Mmio32)
1. 16550 UARTs, which have port IO and MMIO implementations

README.md
jrvanwhy: > I wonder if we should just use this for the example (since nearly everything will use it), and then later introduce the more general version.

I agree, I'll rework this doc to use mmio32_register_layouts/mmio64_register_layouts by default. I will add a section emphasizing that other register layouts are supported as well, though, because I don't want potential users to leave because the README doesn't mention their use case.

jrvanwhy: Rework done. I have a single note at the top that mentions that we support non-MMIO registers, then the rest just uses `mmio32_register_layouts`.

jrvanwhy:
> I would like there to be readmes for the crates which just make it easier for new readers to understand what the different crates are for.

Done, every crate has a README.md now. That might be overkill, but I can delete any that we deem unnecessary.

README.md
bradjc: I don't understand what a register layout is.

codegen/src/parse.rs
jrvanwhy, line 1: # Bikeshed: Punctuation

This bikeshed is entirely about the punctuation in the layout specification syntax, and has no impact on the utility of register_layouts. I just haven't put much thought into the existing design and wanted to throw out some ideas to see what opinions people have. Feel free to add your own ideas. I will add each idea in its own comment so we can thumbs-up and thumbs-down them.

jrvanwhy: # Idea A: Commas, with trailing comma optional

This is the current design:

```rust
mmio32_register_layouts! {
    // Single register definitions
    a: u8 { Read, Write },  // Required comma
    b: a,                   // Required comma

    // Register block definition
    c {
        0 => scalar_definition: u8 { Read, UnsafeWrite },    // Required comma
        1 => array_definition: [[u8; 2]; 3] { UnsafeRead },  // Required comma
        7 => _: 1,                                           // Required comma
        8 => scalar_reference: a,                            // Required comma
        9 => array_reference: [[b; 2]; 3],                   // Optional comma
    },  // Optional comma
}
```

jrvanwhy: # Idea B: Required semicolons on single layouts, no punctuation on register blocks

```rust
mmio32_register_layouts! {
    // Register block definition
    c {
        0 => scalar_definition: u8 { Read, UnsafeWrite },    // Required comma
        1 => array_definition: [[u8; 2]; 3] { UnsafeRead },  // Required comma
        7 => _: 1,                                           // Required comma
        8 => scalar_reference: a,                            // Required comma
        9 => array_reference: [[b; 2]; 3],                   // Optional comma
    }  // No ending punctuation on register blocks

    // Single register definitions
    a: u8 { Read, Write };  // Required semicolon
    b: a;                   // Required semicolon
}
```

jrvanwhy: # Idea C: Required semicolons on single register references, no punctuation on definitions

```rust
mmio32_register_layouts! {
    // Register block definition
    c {
        0 => scalar_definition: u8 { Read, UnsafeWrite },    // Required comma
        1 => array_definition: [[u8; 2]; 3] { UnsafeRead },  // Required comma
        7 => _: 1,                                           // Required comma
        8 => scalar_reference: a,                            // Required comma
        9 => array_reference: [[b; 2]; 3],                   // Optional comma
    }  // No ending punctuation on register blocks

    // Single register definitions
    a: u8 { Read, Write }  // No ending punctuation allowed
    b: a;                  // Required semicolon
}
```

README.md
brghena: I think Brad's proposing `register_layout!`, since the macro just defines a layout for registers for a single peripheral. I'm ambivalent about whether it's `layouts` or `layout`.

bradjc: If it's the layout of bits in a register, this isn't defining that. If it's the layout of registers for a peripheral, which I think it is, that's singular.

jrvanwhy: It is the layout of registers for a peripheral. However, you can define multiple layouts in a single `register_layouts!` invocation, just as you could define multiple structs in a single `register_structs!` invocation.

But I see where that could cause confusion, so maybe `register_layout!` is the better name. I'm also fine with a rename.

bradjc: That's never been done, right? I'd say we've learned we don't need to support that, and perhaps more easily we don't need to encourage it.

jrvanwhy: If you have an array of register blocks, e.g. nrf52's [RamPowerRegisters](https://github.com/tock/tock/blob/78c76c7971318b6c5d29755bcffb6e8937ad4ef5/chips/nrf52/src/power.rs#L77), then you need to define multiple layouts (one for the block, and one for the peripheral that contains the array of blocks).

You could define those in two separate macro calls, but that requires duplicating `#[bus]` attributes between the calls.

bradjc: I'm convinced it's ok to have a compound layout. I still think it's more suitable to call that a single layout.

jrvanwhy: Okay, thumbs-up on this if you would prefer a rename to `register_layout!` and thumbs-down if you think it should stay `register_layouts!`.

EDIT: Nevermind, new vote created below.

Separately, any ideas for what to rename `ast::Layout` to? I don't have any (it took me a while to arrive at the current name), so my inclination is to just leave it as-is and have a different meaning for "layout" internally versus externally. See the [Terminology](https://github.com/jrvanwhy/tock-registers/blob/mod-interface/doc/Terminology.md) document for the internal meaning.

brghena: For `ast::Layout`, wwhy do you need to rename it? I think the internal naming is fine, and honestly it really doesn't matter as far fewer people will look at the internals compared to the public interface

jrvanwhy: I'd like to rename it because it would no longer correspond to the same meaning of the word "layout" as `register_layout!`. But agreed we can leave it as-is because it is internal-only.

I realized now that I never discussed the naming of this macro with the broader Tock community, only with the tock-registers task force. So I think I'll give it a week for people to provide feedback and we can aim to make a final decision at next week's core call.

jrvanwhy: We had a long discussion about naming on the core WG call today and it did not lead to consensus.

One thing we discussed heavily was the option of removing support for multiple Layouts (I'm using the [Terminology](https://github.com/jrvanwhy/tock-registers/blob/mod-interface/doc/Terminology.md)/AST definition here) and instead require separate macro invocations for each Layout. That would simplify things in a few ways:

1. We would no longer have a plural vs. singular debate.
1. We could remove some code from the parsing logic, and it simplifies the AST a bit.
1. I think (not verified) this would allow us to remove Layout's doc comments, because the user could just put the doc comments on the macro invocation instead.
1. It completely removes the [discussion](https://github.com/tock/tock-registers/pull/11#discussion_r3274627305) about separation punctuation because that punctuation would no longer exist.

Of course, the downside of that is that we would not be able to share attributes between separate Layouts. That's probably not a huge deal, because we currently don't have any particularly large attributes that we'd like to share, and a user could always define their own `macro_rules!` macro that wraps our macro and provides the shared attributes.

This discussion resulted in several more name ideas, which I will post below so we can thumbs-up vs. thumbs-down them.

jrvanwhy: Option A: Keep multiple Layout support, rename to `regs!`.

jrvanwhy: Option B: Remove multiple Layout support, rename to `regs!`/`mmio32_regs!`/`mmio64_regs!`.

jrvanwhy: Option C: Remove multiple Layout support, rename to `register_layout!`/`mmio32_register_layout!`/`mmio64_register_layout!`.

jrvanwhy: Option D: Keep multiple Layout support, rename to `register_map!`/`mmio32_register_map!`/`mmio64_register_map!`.

jrvanwhy: Option E: Remove multiple Layout support, rename to `register_map!`/`mmio32_register_map!`/`mmio64_register_map!`.

jrvanwhy: Option F: Keep multiple Layout support, rename to `register_layout!`/`mmio32_register_layout!`/`mmio64_register_layout!`.

jrvanwhy: Option G: Leave it as-is (`register_layouts!`/`mmio32_register_layouts!`/`mmio64_register_layouts!`).

codegen/src/parse.rs
jrvanwhy: Note: this whole bikeshed is moot if we [decide](https://github.com/tock/tock-registers/pull/11#discussion_r3275870143) to remove support for multiple layouts, as there would no longer be any reason to have the punctuation in question.

jrvanwhy: This was discussed briefly on last week's core call. Someone brought up the point that using commas everywhere means there is less for the developer to remember. I agree with that, so at this point no-one has expressed a remaining preference for anything other than idea A. So I'm planning to leave it as-is.

README.md
jrvanwhy: Okay, over Matrix we have decided to:

1. Rename the macro to `register_map!`.
1. Retain multiple Layout support, but minimize its use in the public-facing documentation.

jrvanwhy: Rename and documentation changes complete.

jrvanwhy:
I have performed a second set of renames (recording this here so the PR discussion makes sense in the future):

1. `tock_registers::register_layouts!` is now `tock_registers::register_map!`
1. `tock_registers::mmio32_register_layouts!` is now `tock_registers::mmio32_register_map!`
1. `tock_registers::mmio64_register_layouts!` is now `tock_registers::mmio64_register_map!`

src/fake_register.rs
jrvanwhy: I realized that when you call Read::extract on a FakeRegister, `<FakeRegister as Read>::get` extracts the value from the LocalRegisterCopy, then Read::extract adds it back. That doesn't seem ideal. We can fix that by having FakeRegister implement Read::extract itself, but that feels very much like a case of "you're holding it wrong".

So instead, I think the right solution is to do the simple, obvious thing and have FakeRegister use DT::Value internally rather than LocalRegisterCopy. So I have made that change.

I'm still not 100% confident this is the right move, but given the lack of any other input for over two months and my own uncertainty, I am resolving this comment thread with the hopes of not thinking about this again.

PR #12, opened 2026-04-15
Make tock-registers a Cargo workspace, add a proc-macro crate.

jrvanwhy:
Although we have not decided on a design for the new sound tock-registers interface, I am confident that it will include at least one procedural macro. Procedural macros have to go in their own crate. This adds a new `tock-registers-macros` proc-macro crate, and changes the root Cargo.toml to be a workspace. I also updated the Makefile accordingly, and moved the `miri` invocation out to a separate action (so it is easier to invoke on its own).

I also updated `.lcignore` to ignore all Markdown files, as I am also confident that we will want more documentation files in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 
 .PHONY: test
 test:
-	+RUSTFLAGS="-D warnings" cargo build --no-default-features --workspace
+	+RUSTFLAGS="-D warnings" cargo build --no-default-features
 	+RUSTFLAGS="-D warnings" cargo build --all-targets --workspace
 	+RUSTFLAGS="-D warnings" cargo test --all-targets --workspace
 	+RUSTFLAGS="-D warnings" cargo clippy --workspace
--- a/Makefile
+++ b/Makefile
@@ -2,15 +2,21 @@
 # SPDX-License-Identifier: Apache-2.0 OR MIT
 # Copyright Tock Contributors 2026.
 
+# The Miri test takes longer than all the other tests combined, and is mostly
+# sequential (cargo runs individual test binaries sequentially). Therefore, to
+# speed up `make test`, we split the Miri test and other tests into separate
+# targets (miri_test and fast_tests) so they run in parallel.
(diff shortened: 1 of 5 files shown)


brghena (commented):
Generally I approve here, but with an open question: what is the opinion of procedural macros in Tock right now? Flux currently relies on the `proc-macro2` but is notably an optional dependency for testing. Does this need to be escalated to a decision by the Core team, or have we already decided on it?

jrvanwhy:
> Generally I approve here, but with an open question: what is the opinion of procedural macros in Tock right now? Flux currently relies on the `proc-macro2` but is notably an optional dependency for testing. Does this need to be escalated to a decision by the Core team, or have we already decided on it?

This probably should be escalated to the core WG. Ultimately, we don't have anything against "add a procedural macro", but in practice any complex procedural macro is going to depend on the `syn`, `quote`, and `proc-macro2` crates, which is currently against the third party dependency policy. Regardless of which design we go with, if we want to support easily testing driver code, then this macro will be complex enough to effectively require those crates.

Makefile
```diff
@@ -4,11 +4,15 @@
 
 .PHONY: test
 test:
-	+RUSTFLAGS="-D warnings" cargo build --all-targets --no-default-features
```
jrvanwhy, line 7: Note: removing `--all-targets` is intentional here. Many of the crate's examples and tests will require the `register_types` default feature and will therefore fail to build with --no-default-features.

jrvanwhy:
Dropping this into draft status. The core WG consensus today was to look into creating an external binary that can perform the proc macro's code expansion, which requires a reorganization of the crates in this PR. I am currently testing that idea, and until we know whether it will work out it would be better to leave this PR unmerged.

jrvanwhy:
Okay, then proc-macro reorganization I'm performing #11 have shaken the Makefile and crate organization up enough that I'd rather just close this and open a replacement PR in the future when we know we're going with that design.

PR #17, opened 2026-06-04
Create the `tock-registers-codegen` crate.

jrvanwhy:
`tock-registers-codegen` will be the core of the procedural macro implementation; it will contain the AST, parsing logic, and code generation logic. However, the actual `proc-macro = true` crate will be a separate crate, not added in this PR, to support users who want to be able to build without depending on the `syn`/`quote`/`proc-macro2` crates.

This adds a bit of the implementation in `lib.rs`, but most of the implementation will come in later PRs.

I made the root `Cargo.toml` into a workspace and modified the Makefile to run the tests against all the crates. I changed the Miri test to run in parallel with the non-Miri tests, and added a toolchain initialization action to support expand_macros_test (which will come in a later PR). This parallelization is to keep the runtime of `make -j# test` reasonable once all of `register_map!` has been merged.

AI use: none

### DIFF
--- a/codegen/README.md
+++ b/codegen/README.md
@@ -1,0 +1,14 @@
+# `tock-registers-codegen` crate
+
+[`tock-registers-macros`](../macros) is the procedural macro crate for
+tock-registers. This crate, `tock-registers-codegen`, is a library crate that
+contains the implementation of tock-registers' procedural macros.
+
+Projects can use this crate to run the logic of tock-registers' procedural
+macros in a context other than a procedural macro. For example, the [macro
+expansion tool](../expand_macros) uses this crate to implement the code
+expansion.
+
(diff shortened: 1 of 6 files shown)


bradjc (approved):

codegen/README.md
bradjc, line 1: Can you add the `cargo tree` output?

jrvanwhy: Done. I assume you don't care about the dev dependencies, which are more complex (tock-registers-codegen's doc tests depend on tock-registers), so I used `cargo tree -e no-dev` to omit them. Let me know if you want those included too.

I also `cargo tree` output to `macros/README.md` and `expand_macros/README.md` in PR 11 to match.

alevy (approved):

PR #18, opened 2026-06-04
Add `register_map!` documentation.

jrvanwhy:
Several of the links don't work yet because they link to source code files that don't exist, but I copied them as-is from #11 where they do work.

AI use: none


bradjc (approved):

jrvanwhy:
As a result of feedback in #24, I removed the `ReadWrite` trait. I've updated the docs in this PR accordingly.

lschuermann (approved):

PR #19, opened 2026-06-11
Add the register_map AST.

jrvanwhy:
This also adds the register_definition function, which will be used by the single and block code generation logic. I added a `cargo test --doc --workspace` invocation into `make test` because apparently `cargo test --all-targets` does not run doc tests.

### DIFF
--- a/codegen/src/ast.rs
+++ b/codegen/src/ast.rs
@@ -14,9 +14,10 @@ use quote::quote;
 use std::{ops::Index, slice};
 use syn::{Attribute, Ident, LitInt, Path, Type, TypePath, Visibility};
 
-/// Represents the full input to the register_map! procedural macro. Note that
-/// `tock_registers::register_map!` prepends `$crate` to the input provided by the user, so that
-/// the generated code can refer to tock_registers even if the user has renamed the crate.
+/// Represents the full input to the register_map! procedural macro.
+///
+/// Note that `tock_registers::register_map!` prepends `$crate` to the input provided by the user,
+/// so that the generated code can refer to tock_registers even if the user has renamed the crate.
 /// Therefore, after `tock_registers::register_map!` is expanded, the full input looks like:
 ///
 /// ```ignore
@@ -117,8 +118,10 @@ impl BusAttr {
     }
 }
 
-/// The part of a register that begins after the register's name. For individual register
-/// layouts, this starts with the colon, and in register blocks this begins with the opening brace.
+/// The part of a register that begins after the register's name.
+///
+/// For individual register layouts, this starts with the colon, and in register blocks this begins
+/// with the opening brace.
 ///
 /// ```ignore
 /// # use tock_registers::{Read, Write};
--- a/codegen/src/ast.rs
+++ b/codegen/src/ast.rs
@@ -1,0 +1,303 @@
+// Licensed under the Apache License, Version 2.0 or the MIT License.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// Copyright Tock Contributors 2026.
+// Copyright Better Bytes 2026.
+
+//! The Abstract Syntax Tree for a register_map! invocation.
+
+#![allow(unused)] // TODO: Remove once code generation (block, single) is merged.
+
+// TODO: Remove `ignore` from the doc tests once the macros and Read/Write ops have been merged.
+
+use proc_macro2::TokenStream;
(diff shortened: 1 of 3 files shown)


codegen/src/lib.rs
```diff
@@ -41,8 +46,55 @@ pub enum Env {
     ProcMacro,
 }
 
+/// Generates the Real struct for a register definition (one that has an operations list).
```
alevy:
```suggestion
/// Generates the real struct for a register definition (one that has an operations list).
```

What does "real" struct mean? Is there a fake struct?

codegen/src/ast.rs
```diff
@@ -0,0 +1,303 @@
...
+
+use proc_macro2::TokenStream;
+use quote::quote;
+use std::{ops::Index, slice};
+use syn::{Attribute, Ident, LitInt, Path, Type, TypePath, Visibility};
+
+/// Represents the full input to the register_map! procedural macro. Note that
+/// `tock_registers::register_map!` prepends `$crate` to the input provided by the user, so that
+/// the generated code can refer to tock_registers even if the user has renamed the crate.
+/// Therefore, after `tock_registers::register_map!` is expanded, the full input looks like:
```
alevy:
```suggestion
/// Represents the full input to the register_map! procedural macro.

/// Note that `tock_registers::register_map!` prepends `$crate` to the input
/// provided by the user, so that the generated code can refer to
/// tock_registers even if the user has renamed the crate. Therefore, after
/// `tock_registers::register_map!` is expanded, the full input looks like:
```

codegen/src/ast.rs
```diff
@@ -0,0 +1,303 @@
...
+    pub fn len(&self) -> usize {
+        match self {
+            BusAttr::Bus(_) => 1,
+            BusAttr::Buses(buses) => buses.len(),
+        }
+    }
+}
+
+/// The part of a register that begins after the register's name. For individual register
+/// layouts, this starts with the colon, and in register blocks this begins with the opening brace.
```
alevy:
```suggestion
/// The part of a register that begins after the register's name.
///
/// For individual register layouts, this starts with the colon, and in
/// register blocks this begins with the opening brace.
```

alevy (approved):

codegen/src/lib.rs
jrvanwhy: Okay, so I was pretty sure that I used the term "Real struct" all over in the comments and documentation, and I wrote up a comment about the meaning of that term. While writing the comment, I realized that the terminology is more complex and often-misleading than I thought (I'm glad you asked the question!).

So I grepped the codebase, and found I only used that terminology in 13 places. So my impression that I used it "all over" is incorrect, and it's very changeable! Regardless, my historical note may be interesting so I kept it below.

I don't know if I have the time to do it right now, but I'll prioritize going through those 13 places and figuring out better terminology, then I'll update this PR (and perhaps the other PRs I have open right now) with the new wording.

# "Real struct" history

When I originally designed things, each type of module generated exposed a type called `Real`:

* For single scalar register definitions and register blocks, `Real` is a `struct`.
* For single array register definitions and single register references, `Real` is a type alias to another type.

To refer to that generically, I used the term "Real struct" (with real capitalized). However, there are other types that share similar properties:

* The `real_$name` structs emitted for each register definition field in a register block.
* `tock_registers::RealRegisterArray`.
* The `Element` struct emitted for single register array definitions.

I've used the term "Real struct" in comments the implementation to refer to all of the above.

However, `register_definition` only generates:

1. The `Real` struct for single scalar register definitions.
1. The `Element` struct for single array register definitions.
1. The `real_$name` structs for register definition fields.

So "Real struct" isn't the right terminology to use here.

jrvanwhy: I've removed the "Real struct" terminology throughout PR 11. In most places I replaced it with some variant of "register accessor" (which I use elsewhere too), but in some other places different wording makes sense. `register_definition` in particular now has a more descriptive comment describing exactly which types it covers (using the terminology from Terminology.md).

codegen/src/ast.rs
jrvanwhy: Done.

codegen/src/ast.rs
jrvanwhy: Done.

alevy (approved):

bradjc (approved):

PR #20, opened 2026-06-11
Add the Bus family of traits.

jrvanwhy:
These traits are used to interface with register hardware. They allow for a single peripheral definition to be used on multiple system (such as the 16550 UART, which has port IO and varying MMIO interfaces). They also provide a way for multi-threaded systems to pass register handles between threads.

### DIFF
--- a/src/bus.rs
+++ b/src/bus.rs
@@ -1,0 +1,188 @@
+// Licensed under the Apache License, Version 2.0 or the MIT License.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// Copyright Tock Contributors 2026.
+// Copyright Better Bytes 2026.
+
+//! # Addresses and Buses
+//!
+//! tock-registers supports many different types of registers, including:
+//! 1. MMIO registers
+//! 2. Several types of LiteX registers (these are MMIO, but have padding within the registers that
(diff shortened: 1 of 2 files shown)


src/bus.rs
```diff
@@ -0,0 +1,188 @@
...
+}
+
+/// Address types should implement `Bus<T>` for each primitive type `T` that they support.
+///
+/// # Safety
+/// PADDED_SIZE must be correct, as the generated code relies on PADDED_SIZE to calculate register
+/// offsets.
+pub unsafe trait Bus<T>: Address {
+    /// The size that a value of type T takes in this bus' address space. This exists because LiteX
+    /// buses have intra-register padding for some types.
```
alevy:
```suggestion
    /// The size that a value of type T takes in this bus' address space.
    // This exists because LiteX buses have intra-register padding for
    // some types.
```

alevy (approved):

src/bus.rs
jrvanwhy: Done. I did change the explanation a bit, because I've also been told the 16550 UART has different register layouts on different buses too.

alevy (approved):

bradjc (approved):

PR #24, opened 2026-06-15
Add the safe register_map! operations: Read, Write, and ReadWrite.

jrvanwhy:
These are the successors to the Readable, Writeable, and ReadWriteable traits for use with the `register_map!` macro.

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -26,7 +26,7 @@ pub trait Read: Register {
         field.read(self.get())
     }
 
-    /// Set the raw register value
+    /// Read value of the given field as an enum member
     ///
     /// The [`register_bitfields!`](crate::register_bitfields) macro will
     /// generate an enum containing the various named field variants and
--- a/src/read.rs
+++ b/src/read.rs
@@ -1,0 +1,213 @@
+// Licensed under the Apache License, Version 2.0 or the MIT License.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// Copyright Tock Contributors 2022.
+// Copyright Better Bytes 2026.
+
+use crate::debug::{RegisterDebugInfo, RegisterDebugValue};
+use crate::fields::{Field, FieldValue, TryFromValue};
+#[cfg(feature = "register_types")]
+use crate::{Address, BorrowedBus, Bus};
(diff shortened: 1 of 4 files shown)


alevy (approved):

src/read.rs
```diff
@@ -0,0 +1,213 @@
...
+        self,
+        field: Field<<Self::DataType as DataType>::Value, <Self::DataType as DataType>::LongName>,
+    ) -> <Self::DataType as DataType>::Value
+    where
+        <Self::DataType as DataType>::Value: UIntLike,
+    {
+        field.read(self.get())
+    }
+
+    /// Set the raw register value
```
lschuermann: This wants updating?

src/write.rs
```diff
@@ -0,0 +1,116 @@
...
+        >,
+    ) where
+        <Self::DataType as DataType>::Value: UIntLike,
+    {
+        self.set(field.modify(original.get()));
+    }
+}
+
+/// A register that can be read and written with the same RegisterLongName. This is automatically
+/// implemented for registers that are both Read and Write with the same RegisterLongName.
```
lschuermann: Does the `RegisterLongName` concept still exist in this new design? Either way, I think this comment is a little misleading given that generic type parameter is not mentioned in the trait definition of `ReadWrite`, so maybe we can update this comment about what's expected _semantically_ of the registers for which this trait is implemented (that both read and write operations affect the same semantic register, not two different registers that coincide to be mounted at the same bus address).

lschuermann (changes requested):
LGTM, two minor nits.

src/read.rs
jrvanwhy: Yes. Looks like a copy+paste error that has survived a long time.

src/write.rs
jrvanwhy: Yes, it still exists. I haven't worked with bitfields enough to be comfortable overhauling how `register_bitfields!` works.

However, registers can no longer have different `RegisterLongName`s for reading versus writing (now, aliased registers are defined as two separate registers, each of which has only one bitfield type). So the entire `ReadWrite` trait is now redundant, which I missed when I first implemented `#[aliased]`. I changed this PR to remove that trait entirely, and move the `modify` method into `Write` with a `Self: Read` bound.

PR #25, opened 2026-06-15
Add array-related traits and types.

jrvanwhy:
These traits will be implemented by the generated code for register arrays (as well as fake versions of those arrays).

### DIFF
--- a/src/array.rs
+++ b/src/array.rs
@@ -1,0 +1,164 @@
+// Licensed under the Apache License, Version 2.0 or the MIT License.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// Copyright Tock Contributors 2026.
+// Copyright Better Bytes 2026.
+
+use crate::{internal::RealPhantom, Address, Span};
+use core::marker::PhantomData;
+
+/// Interface for an array of registers (or register blocks, or register arrays). Each register
+/// type should only implement RegisterArray for a single Len.
+pub trait RegisterArray<L: Len>: Copy {
+    /// The type of each element of this array.
+    type Element: Copy;
+
+    // Default implementations of get() and get_unchecked() are provided that depend on each other.
+    // RealArray implementations need to implement at least one of the methods to avoid infinite
+    // recursion. In general, real register implementations (such as RealRegisterArray) will
+    // implement get_unchecked(), while fake implementations (for unit testing) will probably
+    // implement get().
+
+    /// Returns the `index`-th element of this array, or `None` if `index >= L::LEN`.
+    fn get(self, index: usize) -> Option<Self::Element> {
+        if index >= L::LEN {
+            return None;
+        }
+        // Safety: We returned early if `index >= L::LEN`, so because `index` is an integer type we
+        // know that `index < L::LEN`.
+        Some(unsafe { self.get_unchecked(index) })
+    }
+
+    /// Returns the `index`-th element of this array.
+    /// # Safety
+    /// `index` must be less than `L::LEN`
+    // Because this default implementation will only be used in testing environments, it's okay
+    // (and beneficial) for it to have a runtime check.
+    unsafe fn get_unchecked(self, index: usize) -> Self::Element {
+        self.get(index).unwrap_or_else(|| {
+            panic!(
+                "get_unchecked called with out-of-bounds index {index}; len = {}",
+                L::LEN
+            )
+        })
+    }
+}
+
+/// Trait providing the length of a register array.
+///
+/// # Why does this exist?
+/// The obvious way to design `RegisterArray` is to make its length an associated const:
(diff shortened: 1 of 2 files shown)


src/array.rs
```diff
@@ -0,0 +1,164 @@
...
+/// }
+/// ```
+/// the generated `foo::Interface` trait contains a type bound:
+/// ```ignore
+/// trait Interface {
+///     type a: RegisterArray<Element: RegisterArray<Element: RegisterArray<Element:
+///         Register<DataType = u8> + Read + Write>>>;
+/// }
+/// ```
+/// and unfortunately this bound results in a compiler error. As best as I can tell, what happens
```
alevy: > As best as I can tell

Who is "I"? I mean, obviously it's @jrvanwhy, but that's not meaningful to a reader of these docs. Public docs are not a place to discuss background research or debugging.

jrvanwhy: I removed the whole sentence, it basically served as a "don't try to debug this, it is messy" comment which is not productive.

I'm not a huge fan of the fact that I don't fully understand why the compiler error happens, but I don't want to bother compiler engineers with it any more.

alevy (approved):

src/array.rs
```diff
@@ -0,0 +1,162 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// Copyright Tock Contributors 2026.
+// Copyright Better Bytes 2026.
+
+use crate::{internal::RealPhantom, Address, Span};
+use core::marker::PhantomData;
+
+/// Interface for an array of registers (or register blocks, or register arrays). Each register
+/// type should only implement RegisterArray for a single Len.
+pub trait RegisterArray<L: Len>: Copy {
```
lschuermann, line 16: I'm sure you have a good reason to put this behind another trait, but why not just use a const generic here? (Different from your assoc const discussed below)

src/array.rs
```diff
@@ -0,0 +1,162 @@
...
+    // (and beneficial) for it to have a runtime check.
+    unsafe fn get_unchecked(self, index: usize) -> Self::Element {
+        self.get(index).unwrap_or_else(|| {
+            panic!(
+                "get_unchecked called with out-of-bounds index {index}; len = {}",
+                L::LEN
+            )
+        })
+    }
+}
```
lschuermann: I really don't like these default implementations to be defined recursively in terms of each other. I'd rather provide none. I think that there is a reasonable expectation that when you implement all required methods of a trait correctly, it should work. You'd only need to override a default implementation if you want a more performant or richer implementation of the default behavior, but the default behavior should be correct.

If you really want one of these to have a default, I'd say it should be `get_unchecked` using `get`---that is  a correct albeit not an optimal implementation of this method.

src/array.rs
```diff
@@ -0,0 +1,162 @@
...
+/// trait. To do that, we have to make the `RegisterArray` trait generic. Technically, we could
+/// keep that type parameter separate from the length, but combining them makes the design (and the
+/// generated code) simpler.
+pub trait Len {
+    const LEN: usize;
+}
+
+/// Real implementation of RegisterArray.
+// Safety invariant: `address` points to an array of `L::LEN` consecutive `Element` registers.
+pub struct RealRegisterArray<Element: Span, L: Len> {
```
lschuermann, line 103: Don't like `Real` prefixes for such types, but I don't exactly have any better suggestion either. 😅

lschuermann (changes requested):
LGTM generally, nothing major.

src/array.rs
jrvanwhy: That encounters the same error (though it does reduce the chance of you hitting it, because some of the dimensions of the nested array have to match to hit it). I updated the docs to state that both forms error.

src/array.rs
jrvanwhy: I strongly feel that `get_unchecked` should have a default. Otherwise, a bunch of otherwise-safe code will have to implement it (handwritten fake register arrays, fake versions of register arrays from libraries like register-typestate libraries, etc).

But I was very ambivalent on whether `get` should be provided in the first place so I removed the provided implementation.

src/array.rs
jrvanwhy: Agreed. I basically want to have matching trait and struct names for most of the things in this design.

I could make the trait `RegisterArrayInterface` and the struct `RegisterArray`. I don't like that for two reasons:

1. Driver code should generally use the trait rather than the struct. For array fields, RealRegisterArray shouldn't really appear directly in the user code anywhere. So keeping the trait name shorter is better to limit code verbosity.
2. I want to push people towards writing testable code. Making the trait name shorter than the struct name provides that motivation.

lschuermann (approved):

PR #26, opened 2026-06-17
Add code generation for single registers.

jrvanwhy:
This generates code for all layout types except for register blocks, which are in a separate PR.

I added a `#[rustfmt::skip]` to avoid a merge error between this PR and the block generation PR. I didn't realize it in my parse PR, but `rustfmt`'s choice to line up my `// TODO` comments will cause a merge conflict. I'll clean this up as soon as I can do so without causing that conflict.

### DIFF
--- a/codegen/src/lib.rs
+++ b/codegen/src/lib.rs
@@ -15,6 +15,13 @@ mod ast;
 mod parse;
 #[cfg(all(test, not(miri)))]
 mod parse_tests;
+mod single;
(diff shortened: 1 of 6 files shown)


alevy (approved):

lschuermann (approved):
I skimmed this, and haven't found anything obviously wrong so looks good to me. As with all codegen infrastructure, I think this is code we'll want to just test well using unit tests (as included here) and integration tests.

PR #27, opened 2026-06-17
Add register block code generation.

jrvanwhy:
This generates code for register blocks (which will be most uses of `register_map!`), along with unit tests.


alevy (approved):

lschuermann (approved):
Skimmed this, I read through the code in detail a few weeks back. Again, like all codegen, we'll just want extensive testing for this.

PR #29, opened 2026-06-18
Add the `UnsafeRead` and `UnsafeWrite` operations.

jrvanwhy:
These operations can be used with registers that can cause undefined behavior. They are very similar to `Read`/`Write`, with two main differences:

1. They require the user to comply with the hardware's safety requirements.
2. `UnsafeWrite` lacks a `modify` method, because it doesn't cleanly pair with either `Read` or `UnsafeRead` in the same way that `Write` pairs with `Read`.

Safety can be mixed, e.g. a register can be `Read` and `UnsafeWrite`.

### DIFF
--- a/src/unsafe_read.rs
+++ b/src/unsafe_read.rs
@@ -1,0 +1,198 @@
+// Licensed under the Apache License, Version 2.0 or the MIT License.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// Copyright Tock Contributors 2022.
+// Copyright Better Bytes 2026.
+
+use crate::debug::{RegisterDebugInfo, RegisterDebugValue};
+use crate::fields::{Field, FieldValue, TryFromValue};
+use crate::{DataType, LocalRegisterCopy, Register, UIntLike};
+use core::marker::PhantomData;
+
+/// A register that can be read, but which can induce undefined behavior if read at the wrong time.
+///
+/// # Safety
+/// The exact safety invariants of this register are hardware-dependent and should be specified as
+/// part of the documentation for the register layout. For every function in this trait, the user
+/// is responsible for following the hardware's safety invariants.
+pub trait UnsafeRead: Register {
+    /// Get the raw register value
+    unsafe fn get(self) -> <Self::DataType as DataType>::Value;
+
+    /// Read the value of the given field
+    unsafe fn read(
+        self,
+        field: Field<<Self::DataType as DataType>::Value, <Self::DataType as DataType>::LongName>,
+    ) -> <Self::DataType as DataType>::Value
+    where
+        <Self::DataType as DataType>::Value: UIntLike,
+    {
+        // Safety: The caller has complied with this register's safety requirements.
+        field.read(unsafe { self.get() })
+    }
+
+    /// Read value of the given field as an enum member
+    ///
+    /// The [`register_bitfields!`](crate::register_bitfields) macro will
+    /// generate an enum containing the various named field variants and
+    /// implementing the required [`TryFromValue`] trait. It is accessible as
+    /// `$REGISTER_NAME::$FIELD_NAME::Value`.
+    ///
+    /// This method can be useful to symbolically represent read register field
+    /// states throughout the codebase and to enforce exhaustive matches over
+    /// all defined valid register field values.
+    ///
+    /// ## Usage Example
+    ///
+    /// ```rust
+    /// # use tock_registers::interfaces::Readable;
+    /// # use tock_registers::registers::InMemoryRegister;
+    /// # use tock_registers::register_bitfields;
+    /// register_bitfields![u8,
+    ///     EXAMPLEREG [
+    ///         TESTFIELD OFFSET(0) NUMBITS(2) [
+    ///             Foo = 0,
+    ///             Bar = 1,
+    ///             Baz = 2,
+    ///         ],
+    ///     ],
+    /// ];
+    ///
+    /// let reg: InMemoryRegister<u8, EXAMPLEREG::Register> =
+    ///     InMemoryRegister::new(2);
+    ///
+    /// match reg.read_as_enum(EXAMPLEREG::TESTFIELD) {
+    ///     Some(EXAMPLEREG::TESTFIELD::Value::Foo) => "Tock",
+    ///     Some(EXAMPLEREG::TESTFIELD::Value::Bar) => "is",
+    ///     Some(EXAMPLEREG::TESTFIELD::Value::Baz) => "awesome!",
+    ///     None => panic!("boo!"),
+    /// };
+    /// ```
+    unsafe fn read_as_enum<E: TryFromValue<<Self::DataType as DataType>::Value, EnumType = E>>(
+        self,
+        field: Field<<Self::DataType as DataType>::Value, <Self::DataType as DataType>::LongName>,
+    ) -> Option<E>
+    where
+        <Self::DataType as DataType>::Value: UIntLike,
+    {
+        // Safety: The caller has complied with this register's safety requirements.
+        field.read_as_enum(unsafe { self.get() })
+    }
+
+    /// Make a local copy of the register
+    unsafe fn extract(
+        self,
+    ) -> LocalRegisterCopy<
+        <Self::DataType as DataType>::Value,
+        <Self::DataType as DataType>::LongName,
+    > {
+        // Safety: The caller has complied with this register's safety requirements.
+        LocalRegisterCopy::new(unsafe { self.get() })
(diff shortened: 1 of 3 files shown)


src/unsafe_read.rs
```diff
@@ -0,0 +1,198 @@
...
+use crate::fields::{Field, FieldValue, TryFromValue};
+use crate::{DataType, LocalRegisterCopy, Register, UIntLike};
+use core::marker::PhantomData;
+
+/// A register that can be read, but which can induce undefined behavior if read at the wrong time.
+///
+/// # Safety
+/// The exact safety invariants of this register are hardware-dependent and should be specified as
+/// part of the documentation for the register layout. For every function in this trait, the user
+/// is responsible for following the hardware's safety invariants.
```
alevy, line 16: OK, but if this is a `trait`, clients are going to use this generically. The methods are `unsafe`, which implies they may not guarantee soundness, but there should be _some_ invariant then that the client assumes/ensures/depends on!

src/unsafe_write.rs
```diff
@@ -0,0 +1,74 @@
...
+// Copyright Better Bytes 2026.
+
+use crate::{fields::FieldValue, DataType, LocalRegisterCopy, Register, UIntLike};
+
+/// A register that can be written, but which can induce undefined behavior if written incorrectly.
+///
+/// # Safety
+/// The exact safety invariants of this register are hardware-dependent and should be specified as
+/// part of the documentation for the register layout. For every function in this trait, the user
+/// is responsible for following the hardware's safety invariants.
```
alevy, line 13: Ditto

alevy (changes requested):
This looks fine except I think there do need to be some specific explanation of the safety invariants implementors must provide and clients can rely on.

jrvanwhy:
> This looks fine except I think there do need to be some specific explanation of the safety invariants implementors must provide and clients can rely on.

I don't quite understand what you're asking me to do.

The safety requirements do not come from tock-registers, they come from the users of tock-registers. HOPEFULLY the users put a doc comment on their register definition specifying the safety invariants for that register, but if not, there's nothing that tock-registers can do to fix that.

If you have a generic type and all you know about it is that it implements `UnsafeRead` and/or `UnsafeWrite`, you cannot call any methods on it. In practice this isn't an issue: code should actually know what register it is. E.g. if you have:

```rust
mmio32_register_map {
    dma_rng {
        /// Used to start and stop the DMA operation (ignore that we haven't
        /// designed DMA operations yet). Assume that `DmaStart`'s start operation is unsafe.
        ///
        /// # Safety
        /// When calling `DmaStart::start`, the `address` and `len` registers must represent
        /// a writable slice. That slice must remain writable by the hardware until the DMA
        /// operation is complete.
        0 => ctrl: u16 { DmaStart, DmaStop },

        /// Indicates whether a DMA operation is ongoing.
        2 => status: u16 { DmaStatus },

        /// Address of the buffer the random numbers will be written into.
        ///
        /// # Safety
        /// You must not write `address` while a DMA operation is in progress.
        4 => address: *mut u8 { Read, UnsafeWrite },

        /// Length of the buffer the random numbers will be written into.
        ///
        /// # Safety
        /// You must not write `len` while a DMA operation is in progress.
        8 => len: usize { Read, UnsafeWrite },
    }
}
```

then the rustdoc for the `dma_rng::Interface::{ctrl, address, len}` methods will contain the safety invariants. When you write a function that is generic over `<R: dma_rng::Interface>`, it knows the safety invariants for each of the `UnsafeWrite` operations it has access to.

That said, one thing we could improve on is we could allow for doc comments within the operation list. Then you could write:

```rust
mmio32_register_map {
    dma_rng {
        /// Used to start and stop the DMA operation (ignore that we haven't
        /// designed DMA operations yet). Assume that `DmaStart`'s start operation is unsafe.
        0 => ctrl: u16 {
            /// # Safety
            /// When calling `DmaStart::start`, the `address` and `len` registers must represent
            /// a writable slice. That slice must remain writable by the hardware until the DMA
            /// operation is complete.
            DmaStart,

            DmaStop,
        },

        /// Indicates whether a DMA operation is ongoing.
        2 => status: u16 { DmaStatus },

        /// Address of the buffer the random numbers will be written into.
        4 => address: *mut u8 {
            Read,
            /// # Safety
            /// You must not write `address` while a DMA operation is in progress.
            UnsafeWrite,
        },

        /// Length of the buffer the random numbers will be written into.
        8 => len: usize {
            Read,
            /// # Safety
            /// You must not write `len` while a DMA operation is in progress.
            UnsafeWrite,
        },
    }
}
```

and the safety docs would be emitted onto the `impl DmaStart for real_ctrl`, `impl UnsafeWrite for real_address`, and `impl UnsafeWrite for real_len` impls. I confirmed they do show up in the rustdoc there.

But that's a change we cannot make until code generation is merged, because it affects the parsing code, AST, and operations traits, and some (but not all) of that has been merged.

jrvanwhy:
Hmm, there *might* be a different design that avoids the use of the trait entirely. It's a different design than the PR 11 design, so it's probably not worth pursuing at this point. But I'll write it down for future reference.

When emitting a scalar register, we would invoke the operation macro to add methods to the `Interface` trait and `impl Interface for Real` block:

```rust
mmio32_register_map! {
    foo: u8 { /** Doc comment A */ UnsafeRead, /** Doc comment B */ UnsafeWrite },
}
```

would generate:

```rust
mod foo {
    pub trait Interface {
        UnsafeRead!(interface; [/** Doc comment A */], u8,);
        UnsafeWrite!(interface; [/** Doc comment B */], u8,);
    }
    // ...
    impl<B: Bus> Interface for Real<B>
        where
            // I dunno... generating these where bounds is a problem that I don't
            // have an answer to yet. I have some ideas, some really ugly, one
            // maybe not.
    {
        UnsafeRead!(real; [/** Doc comment A */], u8,);
        UnsafeWrite!(real; [/** Doc comment B */], u8,);
    }
}
```

which would expand to (`///` comments are provided by the UnsafeRead/UnsafeWrite macros, `/**` comments are copied from the input):

```rust
mod foo {
    pub trait Interface {
        // Reads this register.
        //
        /** Doc comment A */
        unsafe fn get(self) -> <u8 as tock_registers::DataType>::Value;

        // Plus many provided UnsafeRead methods, each of which duplicates
        // Doc comment A.

        // Writes this register.
        //
        /** Doc comment B */
        unsafe fn set(self, value: <u8 as tock_registers::DataType>::Value);

        // Plus many provided UnsafeWrite methods, each of which duplicates
        // Doc comment A.
    }
    // ...
    impl<B: Bus> Interface for Real<B>
        where
            // dunno
    {
        unsafe fn get(self) -> <u8 as tock_registers::DataType>::Value {
            unsafe { self.address.read() }
        }

        unsafe fn set(self, value: <u8 as tock_registers::DataType>::Value) {
            unsafe { self.address.write(value) };
        }
    }
}
```

Oh, and register blocks would need to emit `Interface` traits for every field, so that code would become somewhat different.

The main difference in driver code is that instead of `use tock_registers::{Read, Write};`, you'd have to `use`-import the `Interface` trait for *every single register you interact with*. Yuck. We could fix *that* by generating a wrapper struct that wraps a type implementing the interface, but that would add more complexity. Also, we'd have to replace FakeRegister with something else, possibly a separate FakeRegister for each generated module.

Okay, now that I've typed that all out and started to realize some of the consequences, I think it is exceptionally complex relative to the benefits it provides.

jrvanwhy:
Actually, lets wait on this PR. I'm not currently sure there are any valid uses for unsafe `Read` and `Write` traits that are backed by non-atomic volatile ops. I'll keep the Zulip conversation going (on why we need atomic volatile ops for DMA) and then re-open this if UnsafeRead/UnsafeWrite still seem useful.

jrvanwhy:
The core issue of "the trait has an `unsafe` method but doesn't know all the safety invariants of that method" is a fundamental drawback of the overall PR 11 design. I don't see an easy way to fix it, though we can improve it a bit.

I'm still not sure whether we'll want UnsafeRead/UnsafeWrite, and if we do want them whether we want them to be atomic or non-atomic. We might decide that we want them while building a safe DMA abstraction (for the address/length fields), but maybe we won't. And I can see arguments for atomicity being either way. And ultimately, we are still waiting on t-opsem to confirm that volatile atomics are a part of Rust's AM.

So for now, I'm closing this, and I plan to remove UnsafeRead/UnsafeWrite from PR 11 entirely to unblock further progress on merging the design. I copied the documentation improvement to https://github.com/tock/tock-registers/issues/35 because I can implement that whether or not we use UnsafeRead/UnsafeWrite.

PR #32, opened 2026-06-23
Add the multi_crate test.

jrvanwhy:
This test verifies that layouts can refer to layouts declared in different crates.


alevy (approved):

bradjc (approved):
It seems like it would be good to have `tests/multi_crate` at the top level. It can be confusing to know if `_test` means "this is a test of this library", "this tests something else", or "this is a tool that helps tock registers users add tests in their code".

jrvanwhy:
> It seems like it would be good to have `tests/multi_crate` at the top level. It can be confusing to know if `_test` means "this is a test of this library", "this tests something else", or "this is a tool that helps tock registers users add tests in their code".

I didn't think Cargo would be happy with that, but I was wrong. I agree, that seems nicer.

Done.

PR #34, opened 2026-06-23
Add the doc_comments and typestate examples.

jrvanwhy:
These examples are mergeable now that the `mmio*_register_map` macros work.


alevy (approved):

examples/doc_comments.rs
```diff
@@ -0,0 +1,57 @@
...
+        1 => _: 1,
+
+        /// Another register block field doc comment.
+        2 => control: u8 { Read, Write },
+    }
+}
+
+mmio32_register_map! {
+    //! Inner doc comment on the register_map! invocation itself. This doc comment will be copied
+    //! onto both `buttons` and `pins`.
```
bradjc: Could this example give an example of what someone would do with this? Is this for commenting multiple register maps? If so, I'm not sure what comment I would write that I would want applied to every register map, and that I wouldn't just copy to each map.

jrvanwhy: To be honest, there's not much utility to this. I implemented the functionality of copying inner attributes onto each Layout because that's useful for the `#[bus]` and `#[buses]` attributes, and when I implemented doc comment parsing I basically thought "why not do the same for that".

The use I can think of is something like adding a link to the relevant datasheet to each register, but you could also put that into the doc comment for the module that contains the `register_map!` call.

I'm also happy to remove support for inner doc comments entirely.

alevy:
I'm agnostic about whether you remove comment support that is probably not useful or keep in the support for replicating comments in a way that may not be used.

examples/doc_comments.rs
jrvanwhy: I decided to remove support for inner doc comments. I've updated the example in this PR accordingly.

I'll actually remove the support in a separate PR (that's a pre-2.0 task not a pre-0.11 task); I opened #39 to track the removal.

ppannuto (approved):

PR #36, opened 2026-06-28
Add the mmio test and generics example.

jrvanwhy:
The mmio test verifies the `Mmio32` and `Mmio64` buses work correctly (which also verifies that offsets are calculated correctly). It is designed to be run in Miri.

The generics example verifies that buses and operations can be generic (and operations can be generic traits or have associated generic types).

### DIFF
--- a/examples/generics.rs
+++ b/examples/generics.rs
@@ -1,0 +1,124 @@
+// Licensed under the Apache License, Version 2.0 or the MIT License.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// Copyright Tock Contributors 2026.
+// Copyright Better Bytes 2026.
+
+//! Demonstrates register_map!'s support for generic buses and operations (and doubles as a test
+//! that register_map!'s generated code compiles in those cases).
+
+use std::fmt::Debug;
+use std::ptr::{read_volatile, NonNull};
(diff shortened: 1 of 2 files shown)


alevy (approved):
Can you explain what it means "it's meant to be run in MIRI?" First, how, then also what precisely does that afford?

jrvanwhy:
> Can you explain what it means "it's meant to be run in MIRI?" First, how, then also what precisely does that afford?

Ah, I didn't put a ton of thought into the PR description.

This is the end-to-end test for the Mmio32/Mmio64 buses. It creates a buffer on the stack that matches the `outer_block` register map. It then points the register abstraction at that buffer, and then interacts with it as if it was a real peripheral. If the Mmio32/Mmio64 buses are computing the register offsets correctly, the test will pass. If they calculate register offsets incorrectly, or do not access the registers in an appropriate manner, then they will fail or exhibit UB. Obviously UB *could* make the test pass incorrectly, but that UB should be caught if you run this test in Miri (which `make test` does).

Though honestly, maybe this should be called something like `offsets.rs`, because in practice it is also the main test for the register block and register array offset logic. There's actually quite a bit of code that this tests that nothing else tests. That said, it has non-offset-related test logic that is specific to the Mmio* buses, so I could go either way.

PR #47, opened 2026-07-22
build: remove force-run qualifier on test recipies

ppannuto:
The prefix `+` forces a recipie command to execute, ignoring dry-run invocation directives (i.e., `make -n` doesn't do what you expect).

We don't need to force command execution for anything, so just remove the `+` prefixes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,24 +24,24 @@ toolchain:
 
 .PHONY: basic_test
 basic_test: toolchain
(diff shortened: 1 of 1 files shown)


jrvanwhy (approved):
Ah oops, I did that while trying to fix issues with cargo trying to connect to the make jobserver. I don't think it worked, but I left it because I didn't think it did any harm.

PR #48, opened 2026-07-29
Refactor Interface to allow peripherals to be owned through smart pointers.

### LINKED ISSUE #42: Support fake peripherals owned via `Rc`/`Arc`.
Currently, I don't think you can use `Rc`/`Arc` to share a fake peripheral between a unit test case and driver impls. That limits the utility of `register_map!`'s unit test support.

tock/tock#4001 supported more complex ownership models via a blanket impl on any type that `Deref`s to the peripheral. That's a big hammer, and may not be as nice in this design (because this design has a larger number of smaller traits). However, we probably need that in order to implement safe DMA APIs that can easily be unit-tested against.

jrvanwhy:
The existing design assumed that fake peripherals would be used through `&` references. This worked great in all the tests and examples I'm made so far, because they're relatively simple and I've stored the fake peripherals on the stack. However, it falls apart for more complex examples. For an upcoming safe DMA example (#49), I need the safe DMA library to expose a function that instantiates a fake peripheral, and the DMA manager that talks to it, then returns both to the unit test case. That requires something like `Rc`, which is not `Copy`.

This change adds a new `Interface` impl into the generated code that looks like:

```rust
impl<T: Deref<Target: Interface>> Interface for T { /* ... */ }
```

However, to make all the lifetimes work out, I had to change the per-field methods to take `self` by reference, and add generic arguments to the per-field types. Also, to make the above Deref work, fake peripherals now need to implement `Interface` on `Self` rather than `&Self`.

I did not change the output for single registers; it was not necessary and I did not want the lifetime complexity to bleed over into that part of the design (there are a lot more traits involved there).

I've verified that the extra indirection added by this change does not increase code size for the drivers ported in tock/tock#4772.

You can review this PR and #49 concurrently; PR 49 builds on top of this to add a safe DMA API example. These PRs are grouped because it justifies the changes in this PR.


bradjc (dismissed):

tests/array.rs
```diff
@@ -87,16 +87,16 @@ struct FakeIncrement<'f> {
...
             .on_read(|i| i.get())
             .on_write(|i, v| i.set(v))
     }
 
-    type counter = FakeRegister<Self, u8, Safe, NoAccess>;
-    fn counter(self) -> Self::counter {
-        FakeRegister::new(self).on_read(|s| {
+    type counter<'s> = FakeRegister<Self, u8, Safe, NoAccess> where Self: 's;
+    fn counter(&self) -> Self::counter<'_> {
+        FakeRegister::new(*self).on_read(|s| {
```
ppannuto:
```suggestion
    type counter<'s> = FakeRegister<&'s Self, u8, Safe, NoAccess> where Self: 's;
    fn counter(&self) -> Self::counter<'_> {
        FakeRegister::new(self).on_read(|s| {
```

Match the other impls here and avoid making a copy of `FakeIncrement` (not wrong I don't think b/c it just holds refs to cells, but fragile if `FakeIncrement` ever holds something that isn't a reference to an interior mutable thing).

_Maybe_ this potential footgun is a case for continuing to think about more friction for `Copy` further down the abstraction hierarchy, but that's a different discussion for a different day.

ppannuto (commented):
Two small things; both nits really, but should probably fix the README typo at minimum

tests/array.rs
jrvanwhy: I'm not a fan of changing this just "to match the other impls". This is a level of indirection we don't need, and I don't want this code to imply that the indirection is necessary.

This handle is not currently unique, and while I think it is *possible* to make it unique I don't think the complexity is worthwhile (unless we go all the way to full typestates, but that's a much bigger project).

README.md
```diff
@@ -66,15 +66,18 @@ mod peripheral {
...
     // Trait representing the peripheral's registers.
-    pub trait Interface: Copy {
-        type cr: Register<DataType = Control::Register> + Read + Write;
-        fn cr(self) -> Self::cr;
-        type s: Register<DataType = Status::Register> + Read;
-        fn s(self) -> Self::s;
+    pub trait Interface {
+        type cr<'s>: Register<DataType = Control::Register> + Read + Write where Self: 'a;
+        fn cr(&self) -> Self::cr<'_>;
+        type s<'s>: Register<DataType = Status::Register> + Read where Self: 'a;
```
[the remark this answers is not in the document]
jrvanwhy: Done.

ppannuto (approved):

doc/UnitTesting.md
```diff
@@ -70,9 +70,9 @@ mod tests {
     struct FakeRng {
         state: Cell<u8>,
     }
-    impl rng::Interface for &FakeRng {
-        type random_byte = FakeRegister<Self, u8, Safe, NoAccess>;
-        fn random_byte(self) -> FakeRegister<Self, u8, Safe, NoAccess> {
+    impl rng::Interface for FakeRng {
+        type random_byte<'s> = FakeRegister<&'s Self, u8, Safe, NoAccess>;
+        fn random_byte(&self) -> FakeRegister<&'_ Self, u8, Safe, NoAccess> {
```
lschuermann: Without having looked to deep into the codegen, this new API taking a `&self` instead of implementing interfaces on reference types makes sense to me and feels more intuitive than the previous one.

lschuermann (approved):

PR #49, opened 2026-07-29
Add a safe DMA API example.

jrvanwhy:
In order to keep this PR's diffs limited to just the safe DMA API, this PR targets PR 48's branch. We should review both PRs, merge PR 48 first, then point this PR to the `main` branch and merge this (I'm currently trying GH's stacked PRs feature, which *should* automate that).

This example models several crates:

1. `safe_dma` The crate exposing the safe DMA API.
2. `chip_unsafe` The unsafe chip crate, which defines the register map and DMA manager.
3. `chip_fake` The fake chip crate, which provides a fake implementation of the peripheral.
4. `chip_safe` The safe chip crate, which implements the driver (and which unit tests the driver).
5. The `main()` crate, which is somewhat-representative of a board crate (it uses the Real instance of the driver).

Both the tests and example are run in Miri to make sure no UB occurs.

This safe DMA API does have some simplifications to keep it reasonably small for the example:

1. It uses a single `UnsafeWrite` operation, whereas a real DMA API might have separate operations for setting the address versus the length (maybe).
2. It only supports mutable buffers, not const buffers.
3. It only supports `'static` buffers.
4. It only supports a single buffer per peripheral.

This models how drivers (and their unit tests) can be implemented in `#![forbid(unsafe_code)]` crates without unsoundness.

There is one major remaining annoyance: the unit tests have to leak their DMA buffers. I have a few ideas for how to improve this:

1. Design a DMA API that supports non-`'static` buffers. This is too complex for an example (requires `Pin` or a similar mechanism), and might not be what we want in Tock anyway.
2. Add a lifetime to the `Interface` trait to indicate "this peripheral can hold a buffer for X lifetime". The Real implementation would set that lifetime to `'static`, but fakes would set it shorter, allowing unit tests to allocate buffers on the stack (if the fakes are on the stack as well).
3. Maybe the safe DMA API can provide a way to work with heap-allocated buffers? This is a bit tricky because the safe DMA crate is `#![no_std]`, but maybe it could expose it under a cargo feature or maybe there can be a separate unittest-support safe DMA crate that exposes that API.

I'll work on those ideas in the future to see if any are worth doing.


examples/safe_dma/safe_dma.rs
```diff
@@ -0,0 +1,254 @@
...
+        $map:ident,
+        // The name of the address register. Must have datatype *mut u8 and be UnsafeWrite.
+        $address:ident,
+        // The name of the length register. Must have datatype usize and be UnsafeWrite.
+        $len:ident,
+        // The name of the enable register. Must have datatype u8 and be DmaEnable.
+        $enable:ident $(,)?
+    ) => {
+        $visibility struct $struct_name<R> {
+            // The DMA buffer that is current in use, or None if no DMA operation is ongoing.
```
ppannuto:
```suggestion
            // The DMA buffer that is currently in use, or None if no DMA operation is ongoing.
```

examples/safe_dma/safe_dma.rs
```diff
@@ -0,0 +1,254 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// Copyright Tock Contributors 2026.
+
+//! Example of a safe DMA abstraction crate. This is obviously simplified relative to a real DMA
+//! crate, e.g. it only supports 'static buffers.
+
+use core::cell::Cell;
+use tock_registers::{Bus, DataType, Register};
+
+pub trait UnsafeWrite: Register {
```
ppannuto:
```suggestion
pub trait DmaConfig: Register {
// or maybe I've convinced myself of this:
pub trait DmaConfig: tock_registers::UnsafeWrite {} + tock_registers::Read{}
// with open question of how / where you put the _method-specific_ safety comments for the `DmaConfig` view of why things are `unsafe` to set/write/modify etc — maybe that's not important if the common theme is just that the unsafe thing is writing to the register, and all the methods ultimately write to the register..., i.e. the answer is this:
// Safety: <Rules for writing to DmaConfig>
pub unsafe trait DmaConfig: tock_registers::UnsafeWrite {} + tock_registers::Read{}
```

I think tl;dr is embodied in the suggestions.

bikeshed/stream-of-conscious:
 - When I was reading the [register map definition](https://github.com/tock/tock-registers/pull/49/changes#diff-28a0d68942f3c0411f941dbda7e46ff7f808de51a02bbf4348a4e8da239692c1R11-R25), my instinct interpreted `DmaEnable` as something special / unique to the DMA interface, but `UnsafeWrite` as something generic that `tock-registers` provided [yes the imports clarify this, but imports are often further from use] — it seems like if this is really about the safety of writing DMA register, and the interface is about enforcing that, the name should reflect it
 - But then I wondered whether there wouldn't be _other_ things that would want to wrap/protect register access in a similar way? i.e., should `UnsafeWrite` be a generic trait from the `tock-registers` API surface that just exports the same methods as `Write` but marks them all `unsafe`
     - Then DMA could come along and just make a subtrait `trait DmaConfg: tock_registers::UnsafeWrite {}` for context clue.
 - But then I wondered if that was a less-useful indirection, since maybe subtraits wouldn't want all of the `Write` methods (maybe they don't all make sense, maybe they shouldn't all be `unsafe`?)
 - Okay, let's actually reason through this for a second, replicating `trait Write` would obligate these methods:
     - `set`: Write whole register using its storage `T` — **hard to imagine not supporting this**
     - `write`: Write specific fields and zero the rest — **akin in spirit to `set`, missing it just harms the ergonomics of the wrapper type**
     - `modify`: Read the register, modify some bits based on `field` parameter, write back new value — **there exist registers that are truly write-only [especially in aliasing scenarios], but that's orthogonal to whether an "unsafe" would be special**
     - `modify_no_read`: Same principle as `modify`, just with a user-provided read
     - ___Conclusion:___ Yes, all of the methods make sense and they should all be `unsafe`
 - Okay, so this should be named `DmaConfigWrite` then? No, not really, the thing can be read (`modify` assumes that), really it should probably also just replicate the safe `Read` interface; there's nothing unsafe in reading DMA registers [not going to enumerate the methods here again, but I don't see a reason any would need to be unsafe]
 - I think I've convinced myself then that the pattern we should demo for users in our examples is to preserve the registers API when making wrapper types — and that we should include the `UnsafeWrite` trait in `tock-registers`
     - I think we have to just manually copy the methods from the safe `Write` trait? But us doing that copy once and being responsible for maintaining the 1:1 method mapping [rather than wrappers copying all the trait methods] feels like a better approach to ensure that wrapper types adhere to the interface.
 - I acknowledge this isn't fundamental to the idea of safe DMA, but I do think the point of this is as an example to demo `TheRightWay (TM)` to wrap registers, so I think this is worth discussing here.

examples/safe_dma/chip_unsafe.rs
```diff
@@ -0,0 +1,29 @@
...
+        /// Address of the DMA buffer.
+        /// # Safety: As required by the safe_dma crate, this must not be modified while a DMA
+        /// operation is ongoing.
+        0 => address: *mut u8 { UnsafeWrite },
+        /// Length of the DMA buffer.
+        /// # Safety: As required by the safe_dma crate, this must not be modified while a DMA
+        /// operation is ongoing.
+        8 => len: usize { UnsafeWrite },
+        /// DMA enable register.
+        16 => enable: u8 { DmaEnable },
```
ppannuto: `DmaEnable` is also `unsafe` to access. I think we should settle on whether `Safety` comments live here at the register map definition, or at the definition of `trait UnsafeWrite; trait DmaEnable` — I lean towards the latter as the trait definition talks more about _how_ the register is actually used

examples/safe_dma/chip_safe.rs
```diff
@@ -0,0 +1,68 @@
...
+
+impl<R: rng::Interface> Rng<R> {
+    /// Starts filling the provided buffer with random data. The existing data in the buffer is
+    /// used to seed the RNG as well.
+    pub fn getrandom_start(&self, buffer: &'static mut [u8]) {
+        self.manager.start(buffer)
+    }
+
+    /// Stops the getrandom operation, returning the filled buffer. If no getrandom operation was
+    /// started, or one is still ongoing, returns None.
```
ppannuto:
```suggestion
    /// Returns the filled buffer after the getrandom operation has finished.
    ///
    /// If no getrandom operation was started, or one is still ongoing, returns None.
```
Rephrase to make it clearer that this method doesn't actually stop an in-flight transaction

examples/safe_dma/main.rs
```diff
@@ -0,0 +1,86 @@
...
+    use chip_unsafe::RngDma;
+    use core::cell::UnsafeCell;
+    use core::mem::{offset_of, replace};
+    use core::ptr::{null_mut, NonNull};
+    use tock_registers::Mmio64;
+
+    // Fake version of the peripheral registers so we can demonstrate how chip_unsafe::RngDma will
+    // be used for real (in a board crate).
+    #[repr(C)]
+    struct Fake {
```
ppannuto:
```suggestion
    // Replicate the MMIO interface of the peripheral registers so we can demonstrate how
    // chip_unsafe::RngDma will be used for real (in a board crate).
    #[repr(C)]
    struct SimulatedMmio {
```

"fake" is a bit overloaded in this example, since it's also the term used for the interface mocks, which are a different concept. We seem pretty committed to "fake" for mocks throughout this crate, so I'm proposing this as a term to make it clearer that this is a slightly different thing happening here.

ppannuto (changes requested):
Sorry this took a while, I needed a moment where I could think more deeply about this.

I mostly think this is good, but have some thoughts on this as the representative example for how to wrap the registers api.

examples/safe_dma/safe_dma.rs
```diff
@@ -0,0 +1,254 @@
...
+                // Safety: This was converted directly from a valid &'static mut [u8].
+                self.buffer.set(Some(buffer));
+            }
+
+            /// Returns the DMA buffer, if the DMA operation has completed.
+            pub fn get_buffer(&self) -> Option<&'static mut [u8]> {
+                if self.registers.$enable().is_running() {
+                    return None;
+                }
+                // Safety: We already checked the DMA operation is ongoing, so by `self.buffer`'s
```
ppannuto:
```suggestion
                // Safety: We already checked the DMA operation is not ongoing, so by `self.buffer`'s
```

examples/safe_dma/chip_fake.rs
```diff
@@ -0,0 +1,84 @@
...
+impl DmaEnable for FakeEnable<'_> {
+    // This fake RNG always finishes its operation instantly, so the driver can never observe it
+    // running.
+    fn is_running(self) -> bool {
+        false
+    }
+
+    unsafe fn start(self) {
+        let buffer: &mut [u8] =
+            unsafe { from_raw_parts_mut(self.0.address.get(), self.0.len.get()) };
```
ppannuto:
```suggestion
            // Safety: This is called from `unsafe fn start()`, whose caller must ensure the
            // address and len fields form a valid, exclusively-accessible `[u8]` slice for
            // the duration of this call — the same precondition imposed on `DmaEnable::start`.
            unsafe { from_raw_parts_mut(self.0.address.get(), self.0.len.get()) };
```

src/mmio.rs
```diff
@@ -34,6 +34,11 @@ macro_rules! mmio_structs {
                 Self(ptr)
             }
 
+            /// Returns the inner address.
```
ppannuto:
```suggestion
            /// Returns the raw pointer value of this MMIO bus address.
            ///
            /// This is intended for use by unsafe hardware abstraction code (e.g. DMA
            /// controllers) that must pass a bare address to hardware. Most callers
            /// should use the typed read/write methods on this bus instead.
            ///
            /// # Safety
            ///
            /// The returned pointer has no lifetime and carries no aliasing guarantees.
            /// The caller is responsible for ensuring that any use of the pointer
            /// respects the memory model for MMIO accesses on this platform.
```

I'm not sure how I feel about the safety comment on a safe method, but I think there's some value in kind of driving home that this isn't a thing most code should probably be accessing.

examples/safe_dma/safe_dma.rs
```diff
@@ -0,0 +1,254 @@
...
+            /// registers.
+            pub unsafe fn new(registers: R) -> Self {
+                Self {
+                    // Safety: This is not Some(_) so no safety invariant applies.
+                    buffer: core::cell::Cell::new(None),
+                    registers,
+                }
+            }
+
+            #[cfg_attr(not(test), allow(dead_code))]
```
ppannuto:
```suggestion

            // `pub` + allow(dead_code) rather than `#[cfg(test)]` so that external test
            // crates (e.g. chip_fake) can call this.
            #[cfg_attr(not(test), allow(dead_code))]
```

examples/safe_dma/safe_dma.rs
```diff
@@ -0,0 +1,254 @@
...
+
+//! Example of a safe DMA abstraction crate. This is obviously simplified relative to a real DMA
+//! crate, e.g. it only supports 'static buffers.
+
+use core::cell::Cell;
+use tock_registers::{Bus, DataType, Register};
+
+pub trait UnsafeWrite: Register {
+    /// # Safety: The safety invariants are hardware-specific, and depend on which register this
+    /// is.
```
ppannuto:
```suggestion
    /// # Safety
    ///
    /// The safety invariants are hardware-specific, and depend on which register this is.
```

examples/safe_dma/safe_dma.rs
```diff
@@ -0,0 +1,254 @@
...
+    };
+    ($($unknown:tt)*) => {};
+}
+
+// Operation for a DMA enable register. Assumes that 0 == not running and 1 == enabled.
+pub trait DmaEnable: Register {
+    /// Performs the fence necessary to let DMA access this buffer, then starts the DMA operation.
+    ///
+    /// # Safety: The address and length registers must point to a buffer which the hardware may
+    /// read and write.
```
ppannuto:
```suggestion
    /// # Safety
    ///
    /// The address and length registers must point to a buffer which the hardware may read
    /// and write.
```

ppannuto (commented):
For completeness, I also asked my robot friend to do a review. Claude mostly found little style nits; here's the ones worth fixing.

examples/safe_dma/safe_dma.rs
jrvanwhy: While designing this, I flip-flopped back and forth on whether to have a generic `UnsafeWrite` trait or DMA-specific operations (`DmaAddress` and `DmaLen`). Ultimately I came to the conclusion "it doesn't really matter, this is just a proof-of-concept" and stopped worrying about it.

But yes, the fact that this is an example is probably justification to think this through more.

If I were implementing a safe DMA crate for real, my initial inclination is to have DmaAddr and DmaLen operations instead of UnsafeWrite. The only reason these are unsafe is to make non-DmaManager code unable to call them. I'm not sure whether I would make the methods `unsafe` or whether I would use a sealed-caller pattern (like `examples/typestate.rs` uses).

Note that I did previously send a PR with an UnsafeWrite trait for tock-registers, but I [closed it because I didn't have enough information to design it](https://github.com/tock/tock-registers/pull/29#issuecomment-4827770376). For example, I wasn't sure whether the operations should be atomic. And in fact, for DMA, setting the address and length are non-atomic operations but the DmaEnable operations *are* atomic.

I could change this to add the separate operations, and update `dma_manager!` to generate the necessary token types, but IMO that would add a lot of lines of boilerplate for not much benefit. Instead, I'm inclined to leave a comment along the lines of:

```rust
// Simplification: this example uses a single UnsafeWrite trait with `unsafe` operations to prevent
// non-DmaManager code from modifying the address and length registers. Instead, a real safe DMA
// crate would probably define separate address and length operations, and might use a sealed-caller
// pattern instead of `unsafe` to enforce that invariant.
```

on top of UnsafeWrite.

examples/safe_dma/chip_unsafe.rs
jrvanwhy: This concern was brought up at #29. In my opinion, they should be on the trait definition wherever possible. That's possible for DmaEnable but *not* for UnsafeWrite, hence why the safety comments are in different places.

If we REALLY want this to be consistent then we could ban traits like UnsafeRead/UnsafeWrite and force anyone using an unsafe register to implement their own operation. That does solve that particular problem, but forces people to write unsafe code that they wouldn't otherwise have to write.

I did have the idea to allow the register-specific safety comment to be moved onto the operation (#35). That would enable the comment to be emitted onto the Real `UnsafeRead`/`UnsafeWrite` impl. I just realized that [it has a different drawback, however](https://github.com/tock/tock-registers/issues/35#issuecomment-5363705049), which might not be worth it.

examples/safe_dma/chip_fake.rs
jrvanwhy: Safety comment added. I had to update DmaEnable's safety invariants as well to require the buffer to be initialized, which seems reasonable for read-write DMA operations.

examples/safe_dma/chip_safe.rs
jrvanwhy: Fixed, that was definitely left over from before my realization that I didn't need to bother with waiting.

examples/safe_dma/main.rs
jrvanwhy: Good point that this overlaps with "fake" elsewhere. Renamed as suggested.

I opened #63 because this made me realize that maybe FakeRegister might be named incorrectly.

examples/safe_dma/safe_dma.rs
jrvanwhy: Done.

examples/safe_dma/safe_dma.rs
jrvanwhy: Done.

examples/safe_dma/safe_dma.rs
jrvanwhy: Fixed.

examples/safe_dma/safe_dma.rs
jrvanwhy: Turns out that isn't required anymore, removed entirely.

I should remember to use `expect(...)` instead of `allow(...)` more, it would've caught this. I keep forgetting it exists.

examples/safe_dma/safe_dma.rs
jrvanwhy: Reworded.

src/mmio.rs
jrvanwhy: Claude's comment missed the mark pretty far here — it appears to conflate the address of the DMA peripheral's register from the address that is passed into the peripheral, and fundamentally miss the purpose of this function.

To be fair to it, I didn't document what this is for. I added a comment now, hopefully I covered everything important. I don't think it needed a safety comment.

examples/safe_dma/safe_dma.rs
ppannuto: I agree with the uncertainty surfaced in the closed `Unsafe[Read|Write]` design, and agree that deferring what goes in the main tock-registers API makes sense until we have a better sense of the use cases for "unsafe" registers.

I like the comment you noted here and think it covers the key concern, seems like a sufficient fix.

src/mmio.rs
ppannuto: 😬 , sorry, I should've noticed that Claude (and by extension me on filtering feedback) conflated "address" here.

examples/safe_dma/main.rs
```diff
@@ -0,0 +1,86 @@
...
+
+//! This example demonstrates how a safe DMA abstraction can be developed on top of
+//! `register_map!`. It is divided into several modules, each of which behaves like a crate:
+//! - [safe_dma] Is the crate exposing the safe DMA API. Depends on tock-registers.
+//! - [chip_unsafe] Is the unsafe chip crate from Tock's crate hierarchy. Depends on the safe DMA
+//! crate.
+//! - [chip_fake] Is a unit testing support crate. Depends on the unsafe chip crate.
+//! - [chip_safe] Is the main chip crate, which contains the bulk of the peripheral's driver. It
+//! depends on chip_unsafe and has a dev-dependency on chip-fake.
+//! - This file (main.rs) represents a board file, instantiating the Real version of the driver.
```
ppannuto: Cosmetic question: Should this be surfaced to a folder-level `README.md` for discoverability? (And I'm unsure if surface means move or copy here, since I imagine that this would be the "first thing seen" in a rustdocs rendering of exploring this example)

ppannuto (dismissed):
Okay, I've been re-reading and re-thinking about this, and I while we need to think more about how we want to handle the UX `unsafe` register operations, I don't think this PR is the right place to debate that design idea (edit: opened #64).

This is a reasonable example for the API as-is, and we shouldn't leave this PR in limbo while bikeshedding exactly where and how safety documentation is done. So I'm going to approve this.

---

That said, I think the comment you suggested should be added for the immediate term, and one hopefully lightweight cosmetic question that would be nice to resolve.

examples/safe_dma/main.rs
jrvanwhy: Yeah, I think a README would be a better place for this.

I decided to move it because I don't really expect people to use rustdoc for their first look at the example (we don't even generate docs for examples right now). I'm not dead set on that however as I don't expect this to change much.

examples/safe_dma/safe_dma.rs
jrvanwhy: Comment added.

examples/safe_dma/chip_unsafe.rs
ppannuto: Moving this discussion to #64

ppannuto (dismissed):

examples/safe_dma/safe_dma.rs
```diff
@@ -0,0 +1,261 @@
+// Licensed under the Apache License, Version 2.0 or the MIT License.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// Copyright Tock Contributors 2026.
+
+//! Example of a safe DMA abstraction crate. This is obviously simplified relative to a real DMA
+//! crate, e.g. it only supports 'static buffers.
```
lschuermann: Maybe this is the origin of my confusion, but I don't understand much of the complexity of this file. What's a "safe DMA abstraction crate", what exactly does it try to abstract, which lower layers does it interact with (and how?), and what high-level interface does it provide?

Adding a bit more prose to explain the purpose of all of the infrastructure of this file could really help here, I think.

examples/safe_dma/safe_dma.rs
```diff
@@ -0,0 +1,261 @@
...
+        }
+        val as u8
+    }
+}
+
+/// Generates the DmaManager type for a peripheral.
+// This is currently hardcoded to a single DMA channel, a real DMA abstraction crate would need to
+// support multiple DMA channels.
+#[macro_export]
+macro_rules! dma_manager {
```
lschuermann: Unless I'm mistaken, we're only calling this macro once for this example, right? Can we simply this to just the bare minimum to demonstrate how tock-register's type can be used to facilitate safe & sound DMA transactions? I feel like right now it's also doing the separate job of showing how an OS like Tock can then build re-usable abstractions that package tock-register's primitives into these higher-level manager structs. To me, that's an important but orthogonal concern. I think this example is trying to do too much.

examples/safe_dma/safe_dma.rs
```diff
@@ -0,0 +1,261 @@
...
+                // safety invariant nothing aliases with `b`. By `self.buffer`'s safety invariant,
+                // `b` meets all other requirements to be converted back to a `&'static mut [u8]`.
+                self.buffer.take().map(|mut b| unsafe { b.as_mut() })
+            }
+        }
+    };
+}
+
+/// A fake register that implements UnsafeWrite by writing the passed value into the given Cell.
+pub struct FakeUnsafeWrite<'c, T: DataType>(&'c Cell<T::Value>);
```
lschuermann: Add some more context, why do we need this?

lschuermann (commented):
The general design here looks right (and sound) to me. I can see how this composes well with such higher-level DMA manager structs, and how it could integrate with constructs like Tock's upstream `DMASlice` to perform other necessary operations like hardware fences.

But this example seems too complex to me. Specifically, IIUC, `safe_dma.rs` defines types like `DMAManager` that are necessary infrastructure for the other modules, but then tries to generalize this into a macro-rules based reusable infrastructure, which makes the code harder to follow.

I think it'd be good if we could simply that file, and add more context and comments because many parts are genuinely non-obvious (even to someone who's spent the better part of the last year looking at DMA soundness).

Then, we could have another example that provides an example of reusable infrastructure to generate such DMA managers. I'm a little skeptical of how much of this is actually tock-register's job, given these managers closely tie in to the driver semantics of the surrounding operating system.

lschuermann:
(Also, my apologies for this review coming so late and blocking this PR)

jrvanwhy:
I believe that most or all of Leon's concerns stem from a failure on my part to communicate the purpose of this example. I've pushed some clarifications to the comments (and the [README](https://github.com/tock/tock-registers/tree/dev/jrvanwhy/safe-dma/examples/safe_dma)). I'll also respond to Leon's comments out-of-order.

> Then, we could have another example that provides an example of reusable infrastructure to generate such DMA managers. I'm a little skeptical of how much of this is actually tock-register's job, given these managers closely tie in to the driver semantics of the surrounding operating system.

I think you're describing *this* example. I'm trying to show how that reusable infrastructure can be built (as well as how it can be used in Tock).

> But this example seems too complex to me. Specifically, IIUC, `safe_dma.rs` defines types like `DMAManager` that are necessary infrastructure for the other modules, but then tries to generalize this into a macro-rules based reusable infrastructure, which makes the code harder to follow.

Showing that DmaManager can be created by reusable infrastructure is a goal here.

> I think it'd be good if we could simply that file, and add more context and comments because many parts are genuinely non-obvious (even to someone who's spent the better part of the last year looking at DMA soundness).

There are comments about the parts I found trickiest — the hardware memory fence and interaction with the atomics model. If there's something else that's tricky that I should comment on, let me know.

> The general design here looks right (and sound) to me. I can see how this composes well with such higher-level DMA manager structs, and how it could integrate with constructs like Tock's upstream `DMASlice` to perform other necessary operations like hardware fences.

Whew, that it looks right/sound is the most important thing at the moment.

examples/safe_dma/safe_dma.rs
jrvanwhy: I don't really see the simplification that is possible here. Are you suggesting I manually define `DmaManager` inside `chip_unsafe.rs`? I guess I could, but it would be almost as verbose and it would make it harder to see which pieces of complexity belong in each crate.

examples/safe_dma/safe_dma.rs
jrvanwhy: Hopefully my README and doc comment changes clarify.

examples/safe_dma/safe_dma.rs
jrvanwhy: I expanded upon FakeUnsafeWrite's doc comment.

lschuermann:
@jrvanwhy Indeed, I totally misunderstood the purpose of this.

I _thought_ that this example was intended to demonstrate that Tock Register's basic primitives are sufficient to implement safe & sound interactions with a DMA peripheral. This would serve two purposes: show, on a fairly low level, what the effects of a given Tock Register's operation w.r.t. the outside world and the Rust Abstract Machine are and how these establish proper fencing and happens-before with a DMA peripheral. And it would implicitly serve as a small, well-defined and constrained Miri test harness, giving both a superficial indicator of soundness and an API drift test.

I now understand that this is actually intending to show _one possible_ way that a downstream OS using Tock Registers _could_ build some reusable abstractions around it. This makes a lot of sense, and in that context, I think this PR is great. My apologies for this confusion.

This PR does most of what my original interpretation of it would do, and more. I'm still of the opinion that a more detailed breakdown of the individual operations, their semantics, and their interactions with the outside world would help, just because of the fact that this example contains a lot of complexity that makes distilling it to its essence harder than necessary. But I don't think this should be a priority; we've got bigger fish to fry and this example accomplishes an important job.

lschuermann (approved):
I'll try to re-review this later today given the new context. But that shouldn't prevent this from being merged, if I don't get to that by end of day today.

lschuermann (approved):
This is great, thank you. Again, sorry for the hold-up!

examples/safe_dma/safe_dma.rs
lschuermann: This was related to my high-level confusion about the purpose of this example and has since been resolved.

PR #56, opened 2026-08-10
codegen: document `lens` module

ppannuto:
This is mostly just a cosmetic thing, but I was working through my understanding of things here again, noticed that there wasn't any doc comments on the `lens` module, but then decided that was appropriate.

With that in mind, I think we should put the hidden-docs cue in the implementation here that it's an internal thing.

AI Disclosure: Claude wrote the actual commit.

### DIFF
--- a/codegen/src/block.rs
+++ b/codegen/src/block.rs
@@ -38,6 +38,7 @@ pub fn generate(env: Env, tock_registers: &Path, layout: &Layout, fields: &[Fiel
     let name = &layout.name;
     let interface_comment = interface_doc_comment();
     let mut interface_fields = TokenStream::new();
+    let lens_comment = lens_doc_comment();
     let mut len_definitions = TokenStream::new();
     let bus_comment = bus_doc_comment();
     let mut bus_bounds = TokenStream::new();
@@ -202,7 +203,7 @@ pub fn generate(env: Env, tock_registers: &Path, layout: &Layout, fields: &[Fiel
             #interface_comment pub trait Interface: #tock_registers::internal::core::marker::Copy {
                 #interface_fields
             }
-            pub mod lens { #len_definitions }
+            #lens_comment pub mod lens { #len_definitions }
             #bus_comment #[allow(clippy::trait_duplication_in_bounds)]
             pub trait Bus: #tock_registers::Address #bus_bounds + sealed::Bus {
                 const SIZE: usize;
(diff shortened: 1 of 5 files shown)


codegen/src/block.rs
```diff
@@ -195,14 +196,19 @@ pub fn generate(env: Env, tock_registers: &Path, layout: &Layout, fields: &[Fiel
     }
 
     // Step 3: the final quote! call that puts everything together.
+    let lens_module = if len_definitions.is_empty() {
+        TokenStream::new()
+    } else {
+        quote! { #lens_comment pub mod lens { #len_definitions } }
+    };
```
jrvanwhy: My suggestions are:

1. Move `let lens_comment = lens_doc_comment();` into this if statement so it is more local (and only called if `lens` is being emitted).
2. Perhaps also merge `len_definitions` and `lens_module` into a single variable called `lengths`.

Then this section would be:

```rust
if !lengths.is_empty() {
    let lens_comment = lens_doc_comment();
    lengths = quote![#lens_comment pub mod lens { #lengths }];
}
```

codegen/src/block_test_docs.rs
jrvanwhy, line 1: The purpose of this test is to make sure the emitted comments look good, so the full expanded comment should be included rather than using `lens_doc_comment()`.

jrvanwhy (commented):
The doc comment looks great!

I'm not sure I agree with suppressing the module when it is empty, for a couple reasons:

1. If someone builds another macro on top of this, they might want to `use $registers::lengths;` and that will fail.
2. The code generation logic is already way more complex/intricate than I'd like. Adding another layer to the logic (to make it `declare variables -> big loop -> postprocessing -> quote![]`) makes that worse.

I don't care strongly about 1 (we can change that in the future in a backwards-compatible way), but I do feel strongly about 2.

I'm still adding comments on how to better implement it, though, just in case we move forward.

codegen/src/block.rs
```diff
@@ -37,6 +37,7 @@ pub fn generate(env: Env, tock_registers: &Path, layout: &Layout, fields: &[Fiel
     let visibility = &layout.visibility;
     let name = &layout.name;
     let interface_comment = interface_doc_comment();
+    let lens_comment = lens_doc_comment();
     let mut interface_fields = TokenStream::new();
```
jrvanwhy: Oops, thought I left this comment on the first pass. I'm trying to keep these variables in the same order as they are used in the final `quote!` call (to the extent that is possible), so `lens_comment` should be between `interface_fields` and `len_definitions`.


```suggestion
    let mut interface_fields = TokenStream::new();
    let lens_comment = lens_doc_comment();
```

This will conflict with #58 -- feel free to combine the PRs into one if that's easier to manage.

jrvanwhy (dismissed):

jrvanwhy (approved):

PR #57, opened 2026-08-10
Add `unsafe` to register map definitions.

jrvanwhy:
This moves the "this register map is a correct description of the hardware" safety invariant into the `register_map` call, thereby removing that invariant from the register handle constructors. This puts that safety invariant on the code that it applies to.

This is particularly useful in multi-crate use cases, where the `register_map!` invocation may be in a different crate from the constructor call. In that case, asserting the safety invariant in the `register_map!` crate is correct.

I suggest putting a `// Safety` comment that links to the corresponding datasheet and describes exactly which peripheral the register map matches.

### DIFF
--- a/doc/Terminology.md
+++ b/doc/Terminology.md
@@ -36,10 +36,10 @@ Examples:
 ```rust
 mmio32_register_map! {
     // A single register layout
-    a: u8 { Read },
+    unsafe a: u8 { Read },
 
     // A register block layout
-    b {
+    unsafe b {
         // `c` is *not* a layout, it is a field of the register block
         0 => c: u8 { Read, Write },
     },
@@ -48,7 +48,7 @@ mmio32_register_map! {
     // block. This is because register_map! is not aware that the element type
     // is a register block, it merely sees that it uses the single register
     // syntax.
-    d: b,
+    unsafe d: b,
 }
 ```
 
@@ -56,7 +56,7 @@ Each field in a register block can be either a register or padding:
 
 ```rust
 mmio32_register_map! {
-    uart {
+    unsafe uart {
         0 => status: u8 { Read },       // `status` is a field.
         1 => _,                         // padding is a field.
         2 => ctrl: u8 { Read, Write },  // `ctrl` is a field.
@@ -74,9 +74,9 @@ Examples:
 
 ```rust
 mmio32_register_map! {
-    a: u8 { Read },  // definition
-    b: a,            // reference (refers to `a`)
-    c {
+    unsafe a: u8 { Read },  // definition
+    unsafe b: a,            // reference (refers to `a`)
+    unsafe c {
         0 => d: u8 { Read },  // definition
         1 => e: a,            // reference (refers to `a`)
     },
@@ -88,9 +88,9 @@ based on whether its type specification contains an array:
 
 ```rust
 mmio32_register_map! {
-    a: u8 { Read },                // A scalar register
-    b: [u8; 2] { Read },           // An array register
-    c {
+    unsafe a: u8 { Read },         // A scalar register
+    unsafe b: [u8; 2] { Read },    // An array register
+    unsafe c {
         0 => d: u8 { Read },       // A scalar register field
         1 => e: [u8; 2] { Read },  // An array register field
     },
@@ -99,13 +99,13 @@ mmio32_register_map! {
     // "array" is based only on the syntax of the register/field, not based on
     // what it points to. Therefore:
 
-    f: b,                // A scalar register
-    g: [a; 2],           // An array register
-    h {
+    unsafe f: b,         // A scalar register
+    unsafe g: [a; 2],    // An array register
+    unsafe h {
         0 => i: b,       // A scalar register field
         1 => j: [a; 2],  // An array register field
     },
-    k: c,                // A scalar register
+    unsafe k: c,         // A scalar register
 }
 ```
 
@@ -113,13 +113,13 @@ To sum it all up:
 
 ```rust
 mmio32_register_map! {
-    a: u8 { Read },       // A single scalar register definition layout
-    b: a,                 // A single scalar register reference layout
-    c: [u8; 2] { Read },  // A single array register definition layout
-    d: [a; 2],            // A single array register reference layout
+    unsafe a: u8 { Read },       // A single scalar register definition layout
+    unsafe b: a,                 // A single scalar register reference layout
+    unsafe c: [u8; 2] { Read },  // A single array register definition layout
+    unsafe d: [a; 2],            // A single array register reference layout
 
     // A register block layout
-    e {
+    unsafe e {
         0 => f: u8 { Read },       // A scalar register definition field
         1 => g: a,                 // A scalar register reference field
         2 => h: [u8; 2] { Read },  // An array register definition field
--- a/README.md
+++ b/README.md
@@ -25,7 +25,9 @@ use tock_registers::{mmio32_register_map, Mmio32, Read, Write};
 // addresses.
 mmio32_register_map! {
     /// Documentation for `peripheral`.
-    peripheral {
+    ///
+    /// # Safety: This matches peripheral X; datasheet: <URL>.
+    unsafe peripheral {
         // Control register: read-write. `Control` is defined using the
         // `register_bitfields!` macro.
         0x000 => cr: Control::Register { Read, Write },
@@ -52,8 +54,9 @@ mmio32_register_map! {
 }
 ```
 
-The fields (registers and padding) must be in order of increasing offset with no
-gaps.
+Declaring a register map is `unsafe` because it is important (for memory safety)
+that the register map be correct for the hardware. The fields (registers and
+padding) must be in order of increasing offset with no gaps.
 
 The macro generates a module that is designed to be used both on real hardware
 and in a unit test environment. Therefore, the module contains traits and a
@@ -101,7 +104,7 @@ the `pub` keyword just before the module name:
 use tock_registers::{mmio32_register_map, Read};
 
 mmio32_register_map! {
-    pub peripheral {
+    pub unsafe peripheral {
         0 => status: u8 { Read },
     }
 }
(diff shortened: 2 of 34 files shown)


doc/Terminology.md
```diff
@@ -99,27 +99,27 @@ mmio32_register_map! {
...
 ```rust
 mmio32_register_map! {
-    a: u8 { Read },       // A single scalar register definition layout
-    b: a,                 // A single scalar register reference layout
-    c: [u8; 2] { Read },  // A single array register definition layout
-    d: [a; 2],            // A single array register reference layout
+    unsafe a: u8 { Read },       // A single scalar register definition layout
+    unsafe b: a,                 // A single scalar register reference layout
+    unsafe c: [u8; 2] { Read },  // A single array register definition layout
+    unsafe d: [a; 2],            // A single array register reference layout
```
ppannuto, line 119: I'm not sure how I feel about this change. This adds a **lot** of instances of the `unsafe` keyword, and I'm not sure the benefit is clear. [n.b., the comment applies to the change as a whole, these were just some convenient lines to highlight as example for discussion]

I feel like what I want to is to assert _one time_ that a register map is accurate. With all of these individual `unsafe`s, I feel an obligation to add a safety comment for _each_ register (`// Safety: This register offset is correct and its type matches the datasheet` times 100...)

My instinct when I see `unsafe` attached to the register definition is that somehow accessing this register could be `unsafe` (e.g., magic register that writes memory somewhere or something) — though I think we were adding the `UnsafeRead`/`UnsafeWrite` types for that elsewhere.

What benefit do we get for making register map authors write the word `unsafe` over and over again? Is there a way we can have one `unsafe`? Or do we even really need `unsafe` in the macro body?

jrvanwhy: When I was writing this, I wasn't entirely sure that I wanted to add `unsafe` to single registers. I did so, but we could also consider adding `unsafe` only to register blocks. That would reduce some of the repetition. There are a couple of bits of awkwardness to that:

1. It makes the syntax for register blocks and single registers more dissimilar, which is awkward.
2. `Span::with_addr` would need to have the "this register layout is correct" safety invariant, because it is used for both register blocks and single registers.

> What benefit do we get for making register map authors write the word `unsafe` over and over again? Is there a way we can have one `unsafe`? Or do we even really need `unsafe` in the macro body?

We [previously decided](https://github.com/tock/tock/blob/master/doc/wg/core/notes/core-notes-2026-05-20.md#renaming-macro) to guide driver authors towards only having one layout per `register_map!` invocation, and to follow that practice in the Tock project. Under that policy, this proposal is equivalent to "one `unsafe` token in each register_map! invocation".

To be clear: I almost tagged this PR "RFC"; I'm totally happy with abandoning this if the consensus is we don't like this.

ppannuto: > `Span::with_addr` would need to have the "this register layout is correct" safety invariant,

I think in my head that's the right place. Ultimately, I don't think it's _unsound_ / unsafe if a layout is wrong when a mock implements the `Interface` (since the layout isn't really used IIUC); it's mapping the layout onto addresses, which the `Real` busses do that carries the safety needs from layout.

I'm trying to make sure I'm imagining how this will be used in practice, since it's likely the chip author that would be validating the safety claims around layout; but I think it's reasonable to encapsulate the call to `with_addr` to the unsafe chip crate; perhaps in a `new` method that validates the address is what's expected, but the address is coming from the board crate as a token that is doing the work of verifying the uniqueness of the MMIO reference?

jrvanwhy: With the existing design, I would expect the encapsulating function to look something like (in `uart.rs`):

```rust
mmio32_register_map![pub registers { ... }];

pub struct UartDriver<R: registers::Interface> { ... }

impl UartDriver<registers::Real<Mmio32>> {
    /// Constructs an instance of the UART on a real chip.
    ///
    /// # Safety:
    /// 1. This is running on chip XYZ.
    pub unsafe fn new_real() -> Self {
        let mmio = Mmio32::from_addr(0x1234);
        // Safety:
        // 1. Chip XYZ's UART is on a 32-bit MMIO bus at address 0x1234.
        // 2. `registers` correctly describes chip XYZ's UART.
        // 3. This is a single-threaded chip with no DMA or other tricky registers,
        //    so no data race is possible.
        let registers = unsafe { registers::Real::new(mmio) };
        Self { registers, ... }
    }
}
```

With this PR, that changes to:

```rust
mmio32_register_map![
    // Safety: This correctly describes chip XYZ's UART.
    unsafe pub registers { ... }];

pub struct UartDriver<R: registers::Interface> { ... }

impl UartDriver<registers::Real<Mmio32>> {
    /// Constructs an instance of the UART on a real chip.
    ///
    /// # Safety:
    /// 1. This is running on chip XYZ.
    pub unsafe fn new_real() -> Self {
        let mmio = Mmio32::from_addr(0x1234);
        // Safety:
        // 1. Chip XYZ's UART is a 0x1234.
        // 2. Chip XYZ's UART is on a 32-bit MMIO bus.
        // 3. This is a single-threaded chip with no DMA or other tricky registers,
        //    so no data race is possible.
        let registers = unsafe { registers::Real::new(mmio) };
        Self { registers, ... }
    }
}
```

So honestly not really different.

In both cases, if we implement something like ChipToken (discussed in Matrix -- effectively a token whose existence asserts "this is chip XYZ"), then these constructors can be safe.

If the peripheral does DMA, then you need uniqueness, but that's a different constraint that this `unsafe` doesn't handle.

ppannuto: I'm a bit torn, but I lean towards the former. [Sorry, this got a bit long / stream of conscious, but hopefully it explains my thinking here]

---

With the new design, there is better separation of `unsafe` concepts and locality of safety rationale. The register map asserts layout and type correctness, rather than deferring that to the hardware `Real` creator to reason about.

On the other hand, I'm not sure that layout and type correctness are actually relevant _Rust safety_ concerns for anything except a `Real` (i.e., I don't think there is anything _unsound_ about an "incorrectly described" peripheral when something other than a `Real` implements the `Interface`).

---

The other thing I'm struggling with is "weirdness" of seeing where/how `unsafe` is dispatched by the macro in the new form.

Usually, when there is something `unsafe`, the imperative code that _uses_ the `unsafe` explains how the safety requirements are upheld. Here, however, it looks like the _definition_ that's making the promise?

Thinking through this a little bit further, if I understand correctly what's happening when everything is expanded, ~~`register_map!` is defining an `unsafe trait` and the created `Real` struct is `impl`ing that unsafe trait~~ — or no, that isn't quite accurate, because it's `**mmio32_**register_map!` which internally also does create the `Real` struct and `impl<B: Bus> Interface for Real<B>` etc.

Maybe that's what is ultimately weird here, the macro _both_ defines an `unsafe trait`, for which I'd expect to see documentation of safety requirements (that, from above, I don't think actually apply to other implementers?) _and_ authors an implementation for the `unsafe trait`, for which I'd expect to see a Safety comment explaining how _this_ impl satisfies the safety requirements. But of course all of that can't be in one macro safety comment. I think this hopefully articulates why the safety-comment-in-macro as shown feels "weird" to me.

---

I _think_ then my question is, if we look at this just as a series of trait definitions, which if any of the traits are `unsafe` from a Rust perspective, and why?

I _think_ I have convinced myself that is it the case that none of the traits are `unsafe`, and it is only the `Real` implementation which is `unsafe` (because it does raw ptr math and dereferences based on the layout and types).

If that is true, then I think `unsafe` should only be attached to `Real` (or any other implementation that has actual, Rust unsafety inside).

jrvanwhy: > On the other hand, I'm not sure that layout and type correctness are actually relevant _Rust safety_ concerns for anything except a `Real` (i.e., I don't think there is anything _unsound_ about an "incorrectly described" peripheral when something other than a `Real` implements the `Interface`).

That's correct, the safety only matters for the real implementation.

> Usually, when there is something `unsafe`, the imperative code that _uses_ the `unsafe` explains how the safety requirements are upheld. Here, however, it looks like the _definition_ that's making the promise?

Yes, the definition is. This is consistent with [unsafe extern blocks](https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-extern.html), where the `unsafe` says "this extern block is correct", but I do see how it's weird relative to other parts of the language (and especially `unsafe` functions).

> Thinking through this a little bit further, if I understand correctly what's happening when everything is expanded, ~`register_map!` is defining an `unsafe trait` and the created `Real` struct is `impl`ing that unsafe trait~ — or no, that isn't quite accurate, because it's `**mmio32_**register_map!` which internally also does create the `Real` struct and `impl<B: Bus> Interface for Real<B>` etc.
> 
> Maybe that's what is ultimately weird here, the macro _both_ defines an `unsafe trait`, for which I'd expect to see documentation of safety requirements (that, from above, I don't think actually apply to other implementers?) _and_ authors an implementation for the `unsafe trait`, for which I'd expect to see a Safety comment explaining how _this_ impl satisfies the safety requirements. But of course all of that can't be in one macro safety comment. I think this hopefully articulates why the safety-comment-in-macro as shown feels "weird" to me.
> 
> I _think_ then my question is, if we look at this just as a series of trait definitions, which if any of the traits are `unsafe` from a Rust perspective, and why?

Interface is a safe trait, not `unsafe`. The Span trait (which is implemented on Real) is unsafe in two ways:

1. It is unsafe to implement, because it is important for offset calculations to be correct.
2. It is unsafe to call Span::with_addr.

> I _think_ I have convinced myself that is it the case that none of the traits are `unsafe`, and it is only the `Real` implementation which is `unsafe` (because it does raw ptr math and dereferences based on the layout and types).
> 
> If that is true, then I think `unsafe` should only be attached to `Real` (or any other implementation that has actual, Rust unsafety inside).

That's a good argument in favor of ditching this PR.

ppannuto: I agree. I'm going to close this PR.

I'm starting work on a branch that tries to make sure Safety comments exist everywhere around `impl Span` and instances of `with_addr`. My primary goal is to make sure I understand how the safety requirements propagate, but if something makes sense out of that work I'll send a PR; likely not today, I'm going to time out soon :/.

jrvanwhy: One other syntax option, require an `unsafe {}` block around the entire input of `mmio32_register_map!`/`mmio64_register_map!`/`register_map!`:

```rust
mmio32_register_map![unsafe { pins: [u32; 4] { Read, Write } }];

mmio32_register_map! { unsafe {
    gpio {
        0 => inputs: pins,
        16 => outputs: pins,
    }
}}
```

PR #58, opened 2026-08-10
rename: Len → Length, lens → lengths

ppannuto:
When I first was reading things, I parsed `lens` in the optics sense, i.e., some kind of view over (part of?) an Interface. That noun stuck in my head even as I understood what the `lens` module was doing; it took me longer than I care to admit to realize that it was the plural of `len` 😬 .

I did a quick search, and while `len` is canonical for length in various Rust things, I can't find any example of other things using plural `len`s, but did find lots of examples of `lens as 'view'`.

This renames `lens` to `lengths` everywhere. If it was just internal, I'd leave it, but as part of the public interface I feel like the more explicit noun is clearer and removes a potential ambiguity.

Claude did the refactor, lovely tool for stuff like this...

This might be annoying churn, and feel free to close/dismiss it if you decide it's on that side of the line.

### DIFF
--- a/codegen/src/block.rs
+++ b/codegen/src/block.rs
@@ -38,7 +38,7 @@ pub fn generate(env: Env, tock_registers: &Path, layout: &Layout, fields: &[Fiel
     let name = &layout.name;
     let interface_comment = interface_doc_comment();
     let mut interface_fields = TokenStream::new();
-    let lens_comment = lens_doc_comment();
+    let lengths_comment = lengths_doc_comment();
     let mut len_definitions = TokenStream::new();
     let bus_comment = bus_doc_comment();
     let mut bus_bounds = TokenStream::new();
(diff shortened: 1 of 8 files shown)


examples/rng_expanded.rs
jrvanwhy, line 1: This file should not have been committed.

jrvanwhy (commented):
I agree with renaming the `lens` module to `lengths`, I fully agree that "lens" is misleading.

I'm more ambivalent about renaming the `Len` trait to `Length`; I don't see how it is confusing, but if it was confusing to you then we should move forward with that.

I don't really agree with expanding the singular "len" into "length" in local variables in `block.rs`; that seems more noisy than useful to me. Though we should resolve https://github.com/tock/tock-registers/pull/56#discussion_r3754325523 before changing that (because that suggestion removes some of those variables entirely).

examples/rng_expanded.rs
ppannuto: UGH, as much as I am enjoying parts of this agentic coding experiment, the baby-sitting is obnoxious; I've had to tell it to remove this file from a stupid number of commits; sorry I missed it here

codegen/src/block.rs
```diff
@@ -134,12 +134,11 @@ pub fn generate(env: Env, tock_registers: &Path, layout: &Layout, fields: &[Fiel
         };
         // Loop that runs once for each level of array nesting.
         for (len_type, size) in len_types_sizes {
-            interface_bound =
-                quote![#tock_registers::RegisterArray<lens::#len_type, Element: #interface_bound>];
+            interface_bound = quote![#tock_registers::RegisterArray<lengths::#len_type, Element: #interface_bound>];
```
jrvanwhy, line 137: I'm not requesting any change, just expressing surprise.

What is `rustfmt` doing here? This PR makes this statement *longer* and now suddenly it belongs on one line?!

jrvanwhy (dismissed):

jrvanwhy (approved):

PR #60, opened 2026-08-13
Mmio32/Mmio64: Panic in systems with the wrong bus width.

### LINKED ISSUE #44: Investigate making operation traits platform-specific
Currently the Bus size calculations for register structs depend on the top-level Interface trait, so register blocks don't compile if the Bus doesn't implement the BusRead/BusWrite/etc traits. That's probably something we can avoid in the generated code, freeing up buses so they don't have to use `unimplemented!()` on arches they don't target.

Should also update the comments in the generated code tests about Interface needing Read/Write bounds, because those bounds would become meaningful.

This does backtrack on the change I made earlier that made Real the default though.

jrvanwhy:
This makes Mmio32 and Mmio64 panic when you:

1. Try to construct them.
2. Try to perform a read and/or write using them.

This is a bit redundant: it shouldn't be possible for code to perform reads/writes if the can't construct the types, but I decided to keep both just in case some code does `unsafe` shenanigans.

This allows the Real types emitted by register_map to still implement their Interface trait while providing some protection against misuse. This also sets an example for other buses: they don't have to work on systems other than their target system.

`src/mmio.rs` relied on invoking MMIO operations on a bus of the wrong width, so I had to redesign it. I combined the 32-bit and 64-bit tests into a single test, which runs on both a 32-bit target and a 64-bit target in Miri.

I plan to send a separate PR that adds an example bus that doesn't implement BusRead/BusWrite at all on non-target platforms, both to demonstrate that is possible and to make sure that keeps working.


src/mmio.rs
```diff
@@ -23,9 +23,18 @@ macro_rules! mmio_structs {
 
         impl $name {
             /// Returns a new MMIO address containing the given pointer.
+            #[cfg_attr(not(target_pointer_width = $width), track_caller)]
```
ppannuto, line 31: The details of this feedback item are from Claude, as my macro-fu isn't quite good enough to validate the claim it's making, but if Claude is right this seems like a real ergonomics issue if the panic points inside the macro implementation.

> `#[track_caller]` chaining doesn't reach user code.
> 
> Both `new()` and `width_check()` get `#[cfg_attr(not(...), track_caller)]`. But `#[track_caller]` only goes one level up: a `panic!` inside `width_check()` will report the line in `new()` where `width_check()` is called, not the user's call to `new()`. The user sees a line inside `mmio.rs`, which is confusing.
>
> The fix is to put the `panic!` directly in the caller rather than in a helper — or to accept that only `width_check`'s frame is tracked. The redundancy the PR describes (checking at both construction and
  read/write) means the tracking issue is multiplied across six call sites.

tests/mmio.rs
```diff
@@ -3,6 +3,10 @@
 // Copyright Tock Contributors 2026.
 // Copyright Better Bytes 2026.
 
+//! This tests verifies that the Mmio32 and Mmio64 buses work correctly. Indirectly, it also
```
ppannuto:
```suggestion
//! These tests verify that the Mmio32 and Mmio64 buses work correctly. Indirectly, it also
```

src/mmio.rs
```diff
@@ -23,9 +23,18 @@ macro_rules! mmio_structs {
...
             pub const fn new(ptr: $storage) -> Self {
+                Self::width_check();
                 Self(ptr)
             }
+
+            /// Panics if this system's bus width does not match this Mmio* type.
+            #[cfg_attr(not(target_pointer_width = $width), track_caller)]
+            const fn width_check() {
+                #[cfg(not(target_pointer_width = $width))]
+                panic!("wrong bus");
```
ppannuto:
```suggestion
                panic!(concat!(stringify!($name), " requires a ", $width, "-bit system"));
```

I _think_ that's valid? Or something that makes a bit more explicit what mismatch happened.

tests/mmio.rs
```diff
@@ -3,6 +3,10 @@
 // Copyright Tock Contributors 2026.
```
ppannuto, line 3: [comment doesn't apply to this line, just somewhere in this file probably]

Should we also add explicit tests that the panics actually fire? i.e.

```rust
  #[test]
  #[cfg(not(target_pointer_width = "32"))]
  #[should_panic(expected = "Mmio32 requires a 32-bit system")]
  fn mmio32_new_panics_on_wrong_width() {
      let _ = Mmio32::new(NonNull::<()>::dangling());
  }

  #[test]
  #[cfg(not(target_pointer_width = "32"))]
  #[should_panic(expected = "Mmio32 requires a 32-bit system")]
  fn mmio32_from_addr_panics_on_wrong_width() {
      let _ = Mmio32::from_addr(0x1000);
  }

  #[test]
  #[cfg(not(target_pointer_width = "64"))]
  #[should_panic(expected = "Mmio64 requires a 64-bit system")]
  fn mmio64_new_panics_on_wrong_width() {
      let _ = Mmio64::new(NonNull::<()>::dangling());
  }

  #[test]
  #[cfg(not(target_pointer_width = "64"))]
  #[should_panic(expected = "Mmio64 requires a 64-bit system")]
  fn mmio64_from_addr_panics_on_wrong_width() {
      let _ = Mmio64::from_addr(0x1000);
  }
```

(these assume the updated language above)

src/mmio.rs
```diff
@@ -23,9 +23,18 @@ macro_rules! mmio_structs {
 
         impl $name {
             /// Returns a new MMIO address containing the given pointer.
```
ppannuto, line 25:
```suggestion
            /// Returns a new MMIO address containing the given pointer.
            ///
            /// # Panics
            /// Panics if this system's pointer width does not match this bus type.
```

src/mmio.rs
```diff
@@ -78,6 +87,7 @@ macro_rules! from_addr_nonnull {
         /// address 0, use one of the nullable MMIO structs.
```
ppannuto, line 95:
```suggestion
        /// address 0, use one of the nullable MMIO structs. Also panics if this system's pointer
        /// width does not match this bus type.
```

ppannuto (changes requested):
This looks reasonable. Mostly small cosmetic / doc requests here.

src/mmio.rs
jrvanwhy: Claude's wrong, `#[track_caller]` is transitive and *should* correctly identify the calling code.

(There is an ounce of truth here... `#[track_caller]` is not *guaranteed* to identify the correct location, but it should in a good compiler as long as you don't do something like call through a non-`#[track_caller]` function pointer).

src/mmio.rs
jrvanwhy: Done.

src/mmio.rs
jrvanwhy: Done.

tests/mmio.rs
jrvanwhy: Tests added.

tests/send_register.rs
jrvanwhy, line 1: I changed the Makefile to run all tests on 32-bit and 64-bit targets in Miri, so I had to update this test to work in both environments.

tests/mmio.rs
jrvanwhy: Fixed.

src/mmio.rs
jrvanwhy: Yeah, that works. I had it in my mind that this is embedded code so I should keep the panic message short, but in retrospect this is only compiled if built for the wrong target architecture anyway so it doesn't matter.

bradjc (dismissed):

ppannuto (approved):

PR #61, opened 2026-08-13
Add an example of a bus that only works on some targets.

### LINKED ISSUE #44: Investigate making operation traits platform-specific

jrvanwhy:
This bus only implements BusRead on the `x86` target. `register_map!` should work with that, and produce Real structs that still exist but which only implement Interface on that target. That currently works, and this example verifies that it will keep working.

Note: treat this PR as stacked on top of #48 (that PR happens to make this work).


bradjc (dismissed):

examples/bus_target_cfg.rs
```diff
@@ -0,0 +1,73 @@
...
+//! can support operations on ONLY the target architecture it is meant for, and serves as a test to
+//! make sure that [register_map]'s generated code does not break that functionality.
+
+use core::marker::PhantomData;
+use tock_registers::{register_map, Address, Bus, Read};
+
+/// A Bus for x86's port I/O. This only implements byte-sized reads to limit the size of this
+/// example.
+#[derive(Clone, Copy)]
+pub struct Port {
```
ppannuto: bikeshed: `PortIo` or `PortIoBus` ?

`Port` is such an overloaded term, and feel like being a bit pedantic to keep examples clear of potential confusion from external context when you're diving into something new has value.

examples/bus_target_cfg.rs
```diff
@@ -0,0 +1,73 @@
...
+    number: u16,
+    _phantom: PhantomData<*mut ()>,
+}
+
+impl Port {
+    /// Returns a new Port with the given number.
+    // This is optional as well, but kept for this example because we use it in main().
```
ppannuto: What's optional, implementing the `new` method at all? Taking a `number`? Maybe just delete this comment? It's already a given that this is "an example of a hardware implementation", so some amount of "this is here because it's the example we're showing" is possibly a given?

examples/bus_target_cfg.rs
```diff
@@ -0,0 +1,73 @@
...
+// environments. Those test depend on the Bus trait, which has Address as a supertrait. Therefore,
+// we have to implement Address regardless of the target.
+impl Address for Port {
+    unsafe fn byte_add(self, offset: usize) -> Port {
+        Port {
+            number: self.number + offset as u16,
+            _phantom: PhantomData,
+        }
+    }
+}
```
ppannuto: I struggled with what this is trying to communicate here, is it saying that this is a non-meaningful thing for Fixed Port peripherals (but mandatory for the interface)? Or just that the words `Address` and `byte_add` are a bit confusing in the context of port-mapped I/O?

```rust
impl Address for FixedPort {
    unsafe fn byte_add(self, offset: usize) -> FixedPort {
        // Offset-based addressing is not valid for fixed port I/O
        assert!(offset == 0);
        self::Self
    }
}
```

ppannuto (changes requested):
Some thoughts — I know the point of this is just to be an example, so you can tell me to buzz off with these comments, but I get lost sometimes with examples when they are tethered to a real thing (x86 Port I/O), but simplify details.

Especially with Port I/O because it falls in the bin of something I know of, but don't _really_ know that well, so when I'm reading a "Port I/O Example", in my mind I'm both solidifying my knowledge of x86 Port I/O and understanding this registers usage example.

examples/bus_target_cfg.rs
jrvanwhy: Good point, I renamed to PortIo.

examples/bus_target_cfg.rs
jrvanwhy: Oh wow that was a really unclear comment. Reworded to:

```rust
// register_map allows the constructor to be #[cfg(target_arch = ...)] as well. However, this
// example calls PortIo::new from main() regardless of the target architecture, so we don't
// demonstrate that cfg in this example.
```

Hopefully that's clear? This comment, and the next one, are just about what parts of a bus implementation you can `#[cfg]` away.

examples/bus_target_cfg.rs
jrvanwhy: Yeah this is really unclear too. I'm trying to explain that you can't put `#[cfg]` on the `impl Address` block, this isn't anything specific to port-mapped I/O.

The new wording is:

```rust
// `register_map!` emits offset tests that run regardless of what target the code is built for,
// including unit test environments. Those test depend on the Bus trait, which has Address as a
// supertrait. Therefore, we cannot put #[cfg(target_arch = ...)] on the Address implementation.
```

Hopefully that's clear?

examples/bus_target_cfg.rs
```diff
@@ -0,0 +1,75 @@
+// Licensed under the Apache License, Version 2.0 or the MIT License.
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+// Copyright Tock Contributors 2025.
```
ppannuto:
```suggestion
// Copyright Better Bytes 2026.
```

ppannuto (changes requested):
The updates clear up everything that confused me, thanks.

Claude noted a copy/paste issue on the copyright block, noting that "all the other 'redesign-era' files have the newer copyright" — it's probably right that that should update, then I think this is good to go.

examples/bus_target_cfg.rs
jrvanwhy: Much of this file was copied from `arch/x86/src/registers/io.rs` in https://github.com/tock/tock/pull/4772. When I copied from that file, I kept the license header intact.

However, looking at it: everything I copied into that PR was written in 2026, so arguably it is more correct to list the copyright as 2026.

I changed the license header to match the headers I used in the PR 11 design as requested.

jrvanwhy:
> Claude noted a copy/paste issue on the copyright block, noting that "all the other 'redesign-era' files have the newer copyright" [...]

Just so no-one is mislead by this: Claude's observation is not entirely true.

jrvanwhy:
The broken CI is unrelated to this PR; the fix is in #67.

ppannuto (approved):

PR #62, opened 2026-08-14
Reorganize the Makefile and CI workflows.

jrvanwhy:
tock-registers has grown to the point where a single `make test` command run by both developers and CI is no longer scaling well. This completely reorganizes the makefile, splitting every `cargo` command into its own action (so Make can parallelize them) and replacing `make test` with four actions:

1. `make test` is now a quick check that takes about a second on my machine.
2. `make full` runs every check the makefile suppports. It takes about 17 seconds on my machine.
3. `make ci_stable` runs every CI check that uses the stable toolchain.
4. `make ci_nightly` runs every CI check that uses the nightly toolchain (currently just Miri, but now it's multiple invocations in parallel).

This expands the tests run under Miri; now all Miri-enabled tests are run on both 32-bit and 64-bit targets. The `tock-registers-codegen` tests are still disabled in Miri, because they take an unreasonable amount of time to run and there's no benefit to running those in Miri.

I discovered that `cargo doc` invalidates Cargo's cache (and vice versa). So alternating `cargo doc` invocations and other commands takes a long time (on the order of 10 seconds per command, versus under a second with a working cache). Therefore, I moved the `cargo doc` target directory to `doc_target` (I could have also called this `target_doc`, but decided on `doc_target` for better tab-completion).

I split the stable-toolchain and nightly-toolchain steps into separate CI jobs so that they can run in parallel. This splits our jobs based on the toolchain they use, which I think is a good tradeoff between minimizing resource wastage while allowing for parallel jobs to keep CI executions fast. The Tock license checker is still a separate job.


ppannuto (dismissed):
This all looks good to me. I just read the diff, didn't directly test anything, but I both trust you did that and have read enough make to feel comfortable

bradjc (approved):

PR #66, opened 2026-09-03
Don't use `byte_add` on pointers to MMIO memory.

### LINKED ISSUE #65: Unsoundness: byte_add requires an allocation, which MMIO may not have
Background: https://github.com/rust-lang/unsafe-code-guidelines/issues/582

Both the `Mmio*` types and the generated code use the `byte_add` method on pointers to calculate pointer offsets. `byte_add` requires the pointer to point to an allocation. Whether or not MMIO peripherals are in typically allocations in Rust is still to-be-decided, but MMIO peripherals that start at address 0 (supported by `Mmio32Nullable`/`Mmio64Nullable`) are NOT in allocations.

We definitely need to change offsetting operations that might operate on null pointers to use an operation that doesn't require an allocation (e.g. `wrapping_byte_add`). Although it currently seems likely that there will be *some* sort of allocation for other MMIO peripherals, the conservative choice now is to change all of our `byte_add` operations to safe functions.

jrvanwhy:
The byte_add method on Rust pointers works within an allocation. There are a few reasons why our MMIO pointers might not point to an allocation:

1. The exact semantics of MMIO memory in Rust haven't been specified yet.
2. Null pointers do not point to an allocation.
3. We create our pointers with no provenance (or at least will once the MSRV is bumped to a verson that has `without_provenance`), and hence they are not derived from pointers that point to an allocation.

So it is unsound to call `byte_add` on these pointers. This PR instead uses `wrapping_byte_add`, which does not have the allocation requirement. This does not result in any change in generated code for the three drivers I ported in tock/tock#4772 (qemu_i486_virt, qemu_rv32_virt, qemu_rv64_virt).

`tests/mmio.rs` did not catch this unsoundness, because it uses pointers that point to a real Rust allocation. This change adds a new test case which exercises offsetting operations on a pointer that lacks provenance. I've verified that this new test case correctly catches the UB when run under Miri. I also split `tests/mmio.rs` into two files: `mmio_nullable.rs` which tests
Mmio32Nullable/Mmio64Nullable, and `mmio_nonnull.rs` which tests Mmio32/Mmio64.

Fixes #65

No AI was used in this PR.

### DIFF
--- a/src/mmio.rs
+++ b/src/mmio.rs
@@ -45,14 +45,6 @@ macro_rules! mmio_structs {
             }
         }
 
-        impl Address for $name {
-            unsafe fn byte_add(self, offset: usize) -> Self {
-                // Safety: The safety requirements of Address::byte_add require self + offset to
-                // remain within this register span and not wrap.
-                Self(unsafe { self.0.byte_add(offset) })
-            }
-        }
-
         // Safety: Mmio* does not expose safe operations to access registers on its own. Instead,
         // it is used through two types:
         // 1. The register accessor structs, which are always !Send + !Sync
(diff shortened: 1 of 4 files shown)


tests/mmio_nullable.rs
jrvanwhy: GitHub did not pick up that `tests/mmio_nullable.rs` is derived from `tests/mmio.rs`. If you want just the diff, here's the output of `diff tests/mmio_nonnull.rs tests/mmio_nullable.rs`:

<details>
<summary>diff output</summary>

```
6,8c6,8
< //! This test verifies that the Mmio32 and Mmio64 buses work correctly. Indirectly, it also
< //! verifies that register offsets are calculated correctly. It is designed to run in Miri on
< //! 32-bit and 64-bit targets.
---
> //! This test verifies that the Mmio32Nullable and Mmio64Nullable buses work
> //! correctly. Indirectly, it also verifies that register offsets are calculated
> //! correctly. It is designed to run in Miri on 32-bit and 64-bit targets.
12,13c12,13
< use core::{cell::UnsafeCell, ptr::NonNull};
< use tock_registers::{register_map, Mmio32, Mmio64, Read, RegisterArray, Write};
---
> use core::cell::UnsafeCell;
> use tock_registers::{register_map, Mmio32Nullable, Mmio64Nullable, Read, RegisterArray, Write};
17c17
<     #![buses(Mmio32, Mmio64)]
---
>     #![buses(Mmio32Nullable, Mmio64Nullable)]
62c62
< type Mmio = Mmio32;
---
> type Mmio = Mmio32Nullable;
64c64
< type Mmio = Mmio64;
---
> type Mmio = Mmio64Nullable;
99c99
<     let mmio = Mmio::new(NonNull::new(peripheral.get()).unwrap().cast());
---
>     let mmio = Mmio::new(peripheral.get().cast());
175c175
<     let mmio = Mmio::new(NonNull::dangling());
---
>     let mmio = Mmio::new(core::ptr::null_mut());
```

</details>

src/mmio.rs
```diff
@@ -77,65 +69,94 @@ mmio_structs! {
...
+        }
+
+        impl Address for $name {
+            unsafe fn byte_add(self, offset: usize) -> Self {
+                let new_ptr = self.0.as_ptr().wrapping_byte_add(offset);
+                // Safety: `self.0` is a NonNull<()> so we know it isn't null.
+                // Address::byte_add's safety invariant requires this offset to
+                // remain within this register span and not wrap. Therefore the
+                // new address is not null.
+                Self(unsafe { NonNull::new_unchecked(new_ptr) })
```
ppannuto: Clarification question on safety documentation. From `src/bus.rs`, we have:

```rust
35:    /// `self` must point into a register span on the correct bus (based on `Self`). The entire
36:    /// range from `self` to the result (inclusive) must be in bounds of that register span. In
37:    /// particular, this must not "wrap around" the address space.
38:    unsafe fn byte_add(self, offset: usize) -> Self;
```

It's not 100% clear to me from that doc and this here who is responsible for verifying that `offset` is valid. Is the _caller_ of `byte_add` responsible, or is the implementation of the unsafe trait method responsible?

[n.b., I don't think there's anything wrong with implementation anywhere, this is just a question of how we use safety docs to express and discharge responsibility for upholding invariants]

ppannuto (changes requested):
Looks good — just one comment on a safety doc.

tests/mmio_nonnull.rs
```diff
@@ -162,3 +164,44 @@ fn mmio() {
...
 }
+
+/// This performs offsetting operations using a base pointer without provenance.
+/// This is an accurate model for real registers (it differs from [mmio] which
+/// uses pointers with provenance to access a stack variable). In Miri, this can
+/// catch UB caused by using offset functions that require the pointers to point
+/// at allocations.
+#[test]
+fn offset() {
+    let mmio = Mmio::new(NonNull::dangling());
```
ppannuto: This note is from Claude, I'm not sure whether it's worth including as a comment here or not:

> The regression test's provenance-free base pointer comes from `NonNull::dangling()`, but the absence of provenance is a std implementation detail rather than a documented guarantee of `dangling()`.

jrvanwhy: Yeah, that's worth fixing. I updated the code to generate the dangling pointer by offsetting a null pointer, which is guaranteed to have no provenance.

src/mmio.rs
jrvanwhy: There's only one invariant we need to uphold: that `new_ptr` is nonnull. The other points were things I was using to prove that new_ptr is nonnull.

I've rewritten this entire comment, maybe its structure is more clear now?

jrvanwhy (commented):
CI fix is in #67.

ppannuto (dismissed):
The updated comments looks good to me.

---

Claude thinks that `examples/generic.rs` should update as well, since real LiteX would be pointing to hardware and have the same issue. It's probably not wrong, but doesn't need to block this PR either.

jrvanwhy:
> The updated comments looks good to me.
> 
> Claude thinks that `examples/generic.rs` should update as well, since real LiteX would be pointing to hardware and have the same issue. It's probably not wrong, but doesn't need to block this PR either.

It's right, yes. Since we're waiting on a second review anyway, I went ahead and fixed that.

ppannuto (approved):

